PR #9, opened 2015-06-21
Support for .Any() and .All() operators

brantburnett:
Completes support for .Any() and .All() operators against both the main document collection and nested arrays inside document objects.  Includes tests which demonstrate their usage.

Note that the changes required to the handling of FromParts may affect ongoing work on Join operations.  Also, this work won't handle Any and All operations which involve joining to other buckets or documents.  That should probably be implemented after Joins rather than before.


jeffrymorris:
Manually merged.

PR #13, opened 2015-06-22
Implemented support for IS NULL and IS NOT NULL

brantburnett:
Modifications
-------------
When N1QlExpressionTreeVisitor encounters an == null or != null operation,
short circuits out and uses the IS NULL and IS NOT NULL operators.

Note
----
This change only supports null on the right side of the equality
comparison, not the left.


jeffrymorris:
@brantburnett -

All merged, thanks and _awesome_ work!

-Jeff

PR #16, opened 2015-06-26
Improve usability of the Explain() method

brantburnett:
Motivation
----------
There was a disconnect in the Explain() method related to its return type.
It was always returning an IQueryable<T> for the entity type at the time
it was called.  For example, bucket.Queryable<Beer>.Explain() would return an
IQueryable<Beer>, therefore indicating that it was returning a list of
Beers rather than an execution plan.

Modifications
-------------
ExplainResultOperator now inherits from
ValueFromSequenceResultOperatorBase, indicating that it is returning a
single value.  QueryExtensions.Explain now returns a dynamic type, and
calls source.Provider.Execute to run the query immediately when called.

Results
-------
Explain() now returns a dynamic object which will contain the execution
plan.  It also executes immediately using ExecuteScalar<dynamic> rather
than deferring execution, much like the Any() or All() methods.

Note that it might be possible to implement a well-known object structure
for the query plan, and return that instead of a dynamic object.  However,
it would require effort to create the object tree, a working knowledge of
the query execution plan structure, and probably some custom JSON
deserializers to choose the correct object types.

### DIFF
--- a/Src/Couchbase.Linq.Tests/BeerSampleTests.cs
+++ b/Src/Couchbase.Linq.Tests/BeerSampleTests.cs
@@ -128,7 +128,6 @@ namespace Couchbase.Linq.Tests
             {
                 using (var bucket = cluster.OpenBucket("beer-sample"))
                 {
-
                     var beers = (from b in bucket.Queryable<Beer>()
                                  where b.Type == "beer"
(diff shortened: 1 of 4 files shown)


jeffrymorris:
@brantburnett -

Awesome, thanks for the PR!

-Jeff

PR #17, opened 2015-06-27
Support for INNER JOIN, LEFT JOIN, INNER UNNEST, and OUTER UNNEST

brantburnett:
Modifications
-------------
Handles join, group join, and additional from clauses in
N1QlQueryModelVisitor.  Throws exceptions for some known cases that are
unsupported.

Results
-------
Basic joins and unnests are now supported.  Cases where a Where clause is
applied to the inner sequence before the join operation are also handled
by moving the where clause to the main query.

Joins must always be made using the N1Ql.Key static method for the inner
key selector.  This is consistent with the N1QL requirement that joins
must use the ON KEYS statement.

Notes
-----
There are probably still edge cases that are not supported, some of which
are not throwing exceptions.

If you apply a Where clause to the inner sequence before a
DefaultIfEmpty() clause, the behavior between N1QL and LINQ is currently
inconsistent.  The generated N1QL query is filtering the Where clause
after the left join, while LINQ is filtering before the join.  Therefore,
it is possible that data on the left side that should be returned with a
null right side is dropped instead.

### DIFF
--- a/Src/Couchbase.Linq.Tests/QueryGeneration/NestTests.cs
+++ b/Src/Couchbase.Linq.Tests/QueryGeneration/NestTests.cs
@@ -1,0 +1,138 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+using Couchbase.Core;
+using Couchbase.Linq.Extensions;
+using Couchbase.Linq.Tests.Documents;
+using Moq;
+using NUnit.Framework;
+
+namespace Couchbase.Linq.Tests.QueryGeneration
+{
+    [TestFixture]
+    public class NestTests : N1QLTestBase
+    {
+        [Test]
+        public void Test_Unnest_Simple()
+        {
+            var mockBucket = new Mock<IBucket>();
+            mockBucket.SetupGet(e => e.Name).Returns("default");
+
+            var query = from brewery in QueryFactory.Queryable<Brewery>(mockBucket.Object)
+                        from address in brewery.Address
+                        select new {name = brewery.Name, address};
+
+            const string expected = "SELECT brewery.name as name, address as address " +
+                "FROM default as brewery "+
+                "INNER UNNEST brewery.address as address";
+
+            var n1QlQuery = CreateN1QlQuery(mockBucket.Object, query.Expression);
+
+            Assert.AreEqual(expected, n1QlQuery);
+        }
+
+        [Test]
+        public void Test_Unnest_Sort()
+        {
+            var mockBucket = new Mock<IBucket>();
+            mockBucket.SetupGet(e => e.Name).Returns("default");
+
+            var query = from brewery in QueryFactory.Queryable<Brewery>(mockBucket.Object)
+                        from address in brewery.Address
+                        orderby address
+                        select new { name = brewery.Name, address };
+
+            const string expected = "SELECT brewery.name as name, address as address " +
+                "FROM default as brewery " +
+                "INNER UNNEST brewery.address as address " +
+                "ORDER BY address ASC";
+
+            var n1QlQuery = CreateN1QlQuery(mockBucket.Object, query.Expression);
+
+            Assert.AreEqual(expected, n1QlQuery);
+        }
+
+        [Test]
+        public void Test_Unnest_Prefiltered()
+        {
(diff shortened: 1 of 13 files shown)


jeffrymorris:
@brantburnett -

I'll review this as soon as I can; probably later this week. 

Thanks for the submission!

-Jeff

Src/Couchbase.Linq.Tests/QueryGeneration/NestTests.cs
```diff
@@ -0,0 +1,138 @@
...
+            Assert.AreEqual(expected, n1QlQuery);
+        }
+
+        [Test]
+        public void Test_Unnest_Sort()
+        {
+            var mockBucket = new Mock<IBucket>();
+            mockBucket.SetupGet(e => e.Name).Returns("default");
+
+            var query = from brewery in QueryFactory.Queryable<Brewery>(mockBucket.Object)
```
jeffrymorris: I am not fond of the syntax, even though I know you don't have much choice given UNNEST has no LINQ equivalent. I am wondering if it's something we could add with Roslyn (new keywords: NEST and UNNEST)?

If we continued on this route, what would NEST look like?

brantburnett: Jeffry,

I agree, I'm not particularly fond of the syntax using double "from"
keywords either.  It is, however, the syntax defined by LINQ for performing
the exact equivalent of the UNNEST operation.  The functional equivalent is
to use .SelectMany().  For reference, you can look at the "Compound From
Clauses" section of this document:
https://msdn.microsoft.com/en-us/library/bb383978.aspx?f=255&MSPPError=-2147217396
.

I have started looking at the NEST operation, and have not found an
equivalent in LINQ.  So I believe it will have to be implemented with
extensions.  I have a proof of concept working as an extension based on
functions, you can see the rough draft here:
https://github.com/brantburnett/Linq2Couchbase/commit/2cb58026ad52b826d00bfdb7c977b6c893117f9f
.

We could also implement a similar extension for UNNEST which simply
replicates the same functionality as the double from clause and SelectMany,
just to make the usage clearer for N1QL users.  But I would recommend
leaving the implementation I suggested in place as well, as it will be more
intuitive to LINQ users.

Brant

On Tue, Jun 30, 2015 at 5:20 PM, Jeffry Morris <notifications@github.com>
wrote:

> In Src/Couchbase.Linq.Tests/QueryGeneration/NestTests.cs
> <https://github.com/couchbaselabs/Linq2Couchbase/pull/17#discussion_r33625130>
> :
>
> > +            const string expected = "SELECT brewery.name as name, address as address " +
> > +                "FROM default as brewery "+
> > +                "INNER UNNEST brewery.address as address";
> > +
> > +            var n1QlQuery = CreateN1QlQuery(mockBucket.Object, query.Expression);
> > +
> > +            Assert.AreEqual(expected, n1QlQuery);
> > +        }
> > +
> > +        [Test]
> > +        public void Test_Unnest_Sort()
> > +        {
> > +            var mockBucket = new Mock<IBucket>();
> > +            mockBucket.SetupGet(e => e.Name).Returns("default");
> > +
> > +            var query = from brewery in QueryFactory.Queryable<Brewery>(mockBucket.Object)
>
> I am not fond of the syntax, even though I know you don't have much choice
> given UNNEST has no LINQ equivalent. I am wondering if it's something we
> could add with Roslyn (new keywords: NEST and UNNEST)?
>
> If we continued on this route, what would NEST look like?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/couchbaselabs/Linq2Couchbase/pull/17/files#r33625130>.
>



-- 
Thanks,
Brant Burnett
CenterEdge Software
(336) 598-5934
http://centeredgesoftware.com

jeffrymorris: @brantburnett -

TBH, I wasn't aware of the double from clauses. Thanks for sharing.

I agree, extension methods are the way to go with NEST until perhaps C# 6.0 and Roslyn, but I am not 100% about that. I also agree to leave the syntax the same for now, this can be handled in documentation.

Also, If you would like to discuss, my email is jeffry.morris at couchbase dot com. Send me an email and we can do a Skype chat.

Thanks!

-Jeff

jeffrymorris:
@brantburnett -

This is merged, thanks for the submission!

-Jeff

PR #18, opened 2015-07-01
Multiple commits to improve operator support

brantburnett:
Note
----
This pull request will have merge conflicts with open pull request #17 due to the move from handling method calls with simple lambda expressions to using IMethodCallTranslator implementations.  If you decide to accept both pulls, just accept one and let me know.  I'd be glad to do the merge cleanup on the other pull request before you accept it.

Also, we'll probably want to look at the way the IMethodCallTranslatorProvider is being regenerated by each N1QlQueryModelVisitor.  It's more efficient than the previous method of regenerating the list of lambda expressions for each N1QlExpressionTreeVisitor, but some kind of caching is in order.  But I thought that caching approach should probably be discussed before implementation, and it may be too early to worry about optimization.


brantburnett:
@jeffrymorris,

I merged the work on joins and unnest into this pull request, resolving the merge conflicts surrounding the method call translator for N1Ql.Key.

Thanks,
Brant

jeffrymorris:
@brantburnett -

Other than the potential licensing issue, this is awesome. You may want to provide your own implementation of the RegistryBase that is behaviorally equivalent, but different.

-Jeff

brantburnett:
@jeffrymorris,

I rewrote the approach to the method call translation system to not use RegistryBase.  I actually like it a bit better, it feels less cumbersome and uses a defined property on the interface instead of just expecting the presence of a named static field.  There's probably some room to make it more generic in the future, but it should do fine for now since we don't have any other need for the logic elsewhere yet.

Brant

jeffrymorris:
@brantburnett 

I like it better too. Thanks for another great submission!

-Jeff

PR #19, opened 2015-07-04
Support for NEST N1QL statements

brantburnett:
Modifications
-------------
Added IN1QlQueryModelVisitor interface to N1QlQueryModelVisitor to publish
methods used by custom clauses.  Implemented new NestClause and
NestExpressionNode classes to represent both INNER NEST and LEFT OUTER
NEST statements.  Implemented VisitNestClause in N1QlQueryModelVisitor to
handle the new clause.  Added LetParts to QueryPartsAggregator to handle
LET statements, which are used if the inner sequence of the Nest is
prefiltered.  Added Nest and LeftOuterNest static methods as an IQueryable
extension, as well as an IEnumerable extension for consistency.

Results
-------
Both INNER NEST and LEFT OUTER NEST are now fully supported via IQueryable
extension methods.  This includes support for prefiltering the inner
sequence being nested.

Note
----
Manipulating the list of keys is not yet supported.  This should tie in
with later work on subqueries, which should allow
N1QlExpressionTreeVisitor to handle filters and manipulations using the
ARRAY statement.

### DIFF
--- a/Src/Couchbase.Linq.Tests/Clauses/NestClauseTests.cs
+++ b/Src/Couchbase.Linq.Tests/Clauses/NestClauseTests.cs
@@ -1,0 +1,204 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Couchbase.Linq.Extensions;
+using Couchbase.Linq.Metadata;
+using Moq;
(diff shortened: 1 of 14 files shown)


jeffrymorris:
@brantburnett 

It's merged now, thanks for the PR!

-Jeff

PR #20, opened 2015-07-15
Support for [] array index accessors in expressions

brantburnett:
Modifications
-------------
Implemented handling for ExpressionType.ArrayIndex.  Added
ListIndexMethodCallTranslator to handle IList based collections.

Modified DefaultMethodCallTranslatorProvider to check for method call
translators on the generic type as well as on any interfaces.  This allows
ListIndexMethodCallTranslator to simply register against IList<>.get_Item
instead of every possible class which might implement IList<>.


jeffrymorris:
@brantburnett -

This is merged, thanks for the contribution!

-Jeff

PR #22, opened 2015-07-31
Support for USE KEYS statement

brantburnett:
Motivation
----------
Add support for `SELECT * FROM bucket USE KEYS ['key1', 'key2']` syntax.
This is particularly important for future work implementing subqueries,
which require the USE KEYS statement if there is a FROM clause.

Modifications
-------------
Implemented UseKeysClause and UseKeysExpressionNode, with support in
N1QlQueryModelVisitor and QueryPartsAggregator.  Created IQueryable and
IEnumerable UseKeys extensions that accept a list of keys to select.
Added appropriate unit tests.

Results
-------
It is possible to query a document or set of documents by key using the
syntax `bucket.Queryable<T>().UseKeys(new[] {"key1", "key2"});`

### DIFF
--- a/Src/Couchbase.Linq.Tests/BeerSampleTests.cs
+++ b/Src/Couchbase.Linq.Tests/BeerSampleTests.cs
@@ -178,6 +178,24 @@ namespace Couchbase.Linq.Tests
         }
 
         [Test]
+        public void UseKeys_SelectDocuments()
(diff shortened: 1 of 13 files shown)


jeffrymorris:
Brant -

Thanks for submitting!

-Jeff

PR #23, opened 2015-08-02
Improved query generation test speed and compatibility

brantburnett:
Motivation
----------
Query generation tests shouldn't require a connection to a cluster on the
localhost in order to complete.  Instead, they should be able to function
using only fakes.

Modifications
-------------
Added IsClusterRequired property to N1QlTestBase.  The cluster will only
be connected during test setup if this property is overriden to return
true.  Removed the dependency on the cluster from
JsonNetMemberNameResolver, instead accepting an IContractResolver as a
constructor parameter.

Results
-------
All tests in the QueryGeneration folder no longer connect to the local
cluster when running.  This allows these tests to initialize and run more
quickly.  The tests can also be successfully run on a system without a
Couchbase cluster.


jeffrymorris:
Brant -

Thanks for the submission!

-Jeff

PR #25, opened 2015-08-08
Escape all N1QL identifiers with tick (`) characters

brantburnett:
Motivation
----------
It is possible for bucket names or property names to conflict with N1QL
keywords.  For example, a property named "key" will conflict with the N1QL
KEY keyword.  Additionally, while unlikely and difficult, it would be
possible to perform a N1QL injection attack by substituting malicious code
in a bucket or property name.

Modifications
-------------
Moved EscapeIdentifier into N1QlHelpers class, and modified to always
escape with ticks instead of only escaping for identifiers with special
characters.  Updated query generation unit tests for new N1QL results.

Results
-------
All N1QL identifiers for bucket names, properties, and aliases are always
escaped with ticks.  All tests are passing, including integration tests
against Couchbase Server.

### DIFF
--- a/Src/Couchbase.Linq/QueryGeneration/N1QlHelpers.cs
+++ b/Src/Couchbase.Linq/QueryGeneration/N1QlHelpers.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Couchbase.Linq.QueryGeneration
+{
+    /// <summary>
+    /// Helpers for N1QL query generation
+    /// </summary>
+    public static class N1QlHelpers
+    {
+
+        /// <summary>
+        ///     Escapes a N1QL identifier using tick (`) characters
+        /// </summary>
+        /// <param name="identifier">The identifier to format</param>
+        /// <returns>An escaped identifier</returns>
+        public static string EscapeIdentifier(string identifier)
+        {
+            if (identifier == null)
+            {
+                throw new ArgumentNullException("identifier");
+            }
+
+            if (identifier.IndexOf('`') >= 0)
+            {
+                // This should not occur, and is primarily in place to prevent N1QL injection attacks
+                // So it isn't performance critical to perform this replace in the StringBuilder
+
+                identifier = identifier.Replace("`", "``");
+            }
+
+            var sb = new StringBuilder(identifier.Length + 2);
+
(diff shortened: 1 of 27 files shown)


Src/Couchbase.Linq/QueryGeneration/N1QlHelpers.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            if (identifier.IndexOf('`') >= 0)
+            {
+                // This should not occur, and is primarily in place to prevent N1QL injection attacks
+                // So it isn't performance critical to perform this replace in the StringBuilder
+
+                identifier = identifier.Replace("`", "``");
+            }
+
+            var sb = new StringBuilder(identifier.Length + 2);
```
jeffrymorris: Wouldn't a simple concat perform better than using StringBuilder?

brantburnett: Good question, let me do some performance comparisons.

Brant

On Saturday, August 8, 2015, Jeffry Morris <notifications@github.com> wrote:

> In Src/Couchbase.Linq/QueryGeneration/N1QlHelpers.cs
> <https://github.com/couchbaselabs/Linq2Couchbase/pull/25#discussion_r36580995>
> :
>
> > +        public static string EscapeIdentifier(string identifier)
> > +        {
> > +            if (identifier == null)
> > +            {
> > +                throw new ArgumentNullException("identifier");
> > +            }
> > +
> > +            if (identifier.IndexOf('`') >= 0)
> > +            {
> > +                // This should not occur, and is primarily in place to prevent N1QL injection attacks
> > +                // So it isn't performance critical to perform this replace in the StringBuilder
> > +
> > +                identifier = identifier.Replace("`", "``");
> > +            }
> > +
> > +            var sb = new StringBuilder(identifier.Length + 2);
>
> Wouldn't a simple concat perform better than using StringBuilder?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/couchbaselabs/Linq2Couchbase/pull/25/files#r36580995>.
>


-- 
Thanks,
Brant Burnett
CenterEdge Software
(336) 598-5934
http://centeredgesoftware.com

brantburnett:
@jeffrymorris ,

You were right about the performance, updated the pull request.

Thanks,
Brant

jeffrymorris:
Brant -

This is merged, Thanks!

-Jeff

PR #26, opened 2015-08-09
Ensure unique extent names in N1QL queries

brantburnett:
Motivation
----------
The current approach of using clause ItemNames as the extent name in
queries does not guarantee uniqueness.  For more information, see
https://github.com/re-motion/Relinq/blob/develop/Core/Clauses/FromClauseBase.cs#L59.
Ensuring unique extent names is especially important when working with
subqueries, where the possibility of extent name conflicts is much higher.

Modifications
-------------
Created `N1QlExtentNameProvider` to generate and cache extent names during
the query generation process.  Extents are numbered starting with
"Extent1".

Created `N1QlQueryGenerationContext` to wrap together a `N1QlExtentNameProvider`,
`IMemberNameResolver`, `MethodCallTranslatorProvider`, and `ParameterAggregator`.
This makes it simpler to pass these context items between the
`N1QlQueryModelVisitor` and the `N1QlExpressionTreeVisitor`.

Modified all points using ItemName properties to instead use
`N1QlExtentNameProvider.GetExtentName`.  Also replaced `GetNewGenName`
with calls to `GetExtentName` and `GetUnlinkedExtentName`.

Updated query generation tests to reflect the new extent names.

Results
-------
All extents have names which are guaranteed to be unique, rather than
using the ItemName property.  All tests are passing.

### DIFF
--- a/Src/Couchbase.Linq.Tests/BucketExtensionTests.cs
+++ b/Src/Couchbase.Linq.Tests/BucketExtensionTests.cs
@@ -28,7 +28,7 @@ namespace Couchbase.Linq.Tests
                             fname = c.FirstName
                         };
 
-                    const string expected = "SELECT `c`.`age` as `age`, `c`.`fname` as `fname` FROM `default` as `c`";
+                    const string expected = "SELECT `Extent1`.`age` as `age`, `Extent1`.`fname` as `fname` FROM `default` as `Extent1`";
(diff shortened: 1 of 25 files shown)


jeffrymorris:
Brant -

This is merged, thanks for the PR! 

-Jeff

PR #33, opened 2015-09-19
Support for GROUP BY, HAVING, and aggregates

brantburnett:
Motivations
-----------
Allow LINQ group clauses and aggregates to generate valid N1QL statements.
Should support simple aggregates directly against a query, or more
advanced aggregates utilizing groups.  Should also support treating where
clauses after the group clause as HAVING statements.

Modifications
-------------
For simple aggregates, added AggregateFunction to QueryPartsAggregator.
This function wraps SelectParts in the select clause.

Added handling to N1QlQueryVisitor for group clauses.  When grouping is
performed in a subquery as part of the MainFromClause, the group is
flattened onto the main query.  Created KeyExpressionTransformer and
MultiKeyExpressionTransformer classes to transform references to the group
keys after the query has been flattened.  When where clauses are
encountered after flattening, they are added to HavingParts instead of
WhereParts on the QueryPartsAggregator.

Introduced a new N1QlQueryType, Aggregate, which returns a simple
statement such as COUNT(*) from the QueryPartsAggregator.  This is used
when ReLinq generates a subquery in the select clause to perform one part
of a grouped aggregate.

For testing, simple aggregates always return immediately and don't provide
a method to capture the query.  Added the BucketQueryExecutorEmulator
class and the CreateQueryable function to support capturing query text
after the call is complete.  Created tests for all new functionality.

Results
-------
LINQ can now be used to generate queries with GROUP BY and HAVING clauses,
as well as the standard COUNT, COUNT distinct, SUM, AVG, MIN, and MAX
aggregate functions.

### DIFF
--- a/Src/Couchbase.Linq.Tests/BeerSampleTests.cs
+++ b/Src/Couchbase.Linq.Tests/BeerSampleTests.cs
@@ -656,7 +656,7 @@ namespace Couchbase.Linq.Tests
         [Test()]
         public void AggregateTests_SimpleAverage()
         {
-            using (var cluster = new Cluster())
+            using (var cluster = new Cluster(TestConfigurations.DefaultConfig()))
             {
                 using (var bucket = cluster.OpenBucket("beer-sample"))
                 {
@@ -670,7 +670,7 @@ namespace Couchbase.Linq.Tests
         [Test()]
         public void AggregateTests_SimpleCount()
         {
-            using (var cluster = new Cluster())
+            using (var cluster = new Cluster(TestConfigurations.DefaultConfig()))
             {
                 using (var bucket = cluster.OpenBucket("beer-sample"))
                 {
@@ -684,7 +684,7 @@ namespace Couchbase.Linq.Tests
         [Test()]
         public void AggregateTests_GroupBy()
         {
-            using (var cluster = new Cluster())
+            using (var cluster = new Cluster(TestConfigurations.DefaultConfig()))
             {
                 using (var bucket = cluster.OpenBucket("beer-sample"))
                 {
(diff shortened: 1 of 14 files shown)


Src/Couchbase.Linq.Tests/BucketQueryExecutorEmulator.cs
```diff
@@ -0,0 +1,72 @@
...
+using Remotion.Linq;
+
+namespace Couchbase.Linq.Tests
+{
+    /// <summary>
+    /// Used to test query generation for result operators that always execute the query immediately.
+    /// This class fakes the result (always returns null or an empty list), but stores the query that was 
+    /// generated in the Query property.
+    /// </summary>
+    public class BucketQueryExecutorEmulator : IQueryExecutor
```
jeffrymorris: Nice!

Src/Couchbase.Linq.Tests/BeerSampleTests.cs
```diff
@@ -652,5 +652,126 @@ orderby brewery.Name
                 }
             }
         }
+
+        [Test()]
+        public void AggregateTests_SimpleAverage()
+        {
+            using (var cluster = new Cluster())
```
jeffrymorris: This should be using (var cluster = new Cluster(TestConfigurations.DefaultConfig() ){..} so that it's easier to run the test off of local host (by changing a single location in the app.config).

brantburnett: Thanks, missed that.  Corrected with additional commit.

jeffrymorris:
Brant -

Great contribution, thanks!

-Jeff

PR #35, opened 2015-09-29
Improved/fixed support for First and Single operators

brantburnett:
Motivations
-----------
First and single operations can be optimized by not selecting more data
than necessary from the server.  They carry an implicit LIMIT 1 or LIMIT 2
when they are used.

Also, the First operator as implemented was throwing an error if there was
more than one result.  It's behavior was the same as Single.

Modifications
-------------
Include LIMIT 1 on all First operations and LIMIT 2 on all Single
operations.  Added query generation and beer-sample integration tests.


jeffrymorris:
Thanks Brant!

PR #38, opened 2015-09-30
LINQ-9: Remove ClusterHelper reference from N1QLQueryModelVistor

jeffrymorris:
Motivation
----------
The Linq project had a hard reference to the ClusterHelper; this commit
removes the reference and instead allows the ClientConfiguration to be
injected as a reference.

Modification
------------
The CLusterHelper reference was removed and instead ClientConfiguration is
injected from the higher-level objects.

Result
------
ClusterHelper is now no longer a required reference; you do not need to
call ClusterHelper.Initialize() before executing a query.

### DIFF
--- a/Src/Couchbase.Linq/Extensions/BucketExtensions.cs
+++ b/Src/Couchbase.Linq/Extensions/BucketExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System.Linq;
+using Couchbase.Configuration.Client;
 using Couchbase.Core;
 using Couchbase.Linq.Filters;
 
@@ -8,7 +9,8 @@ namespace Couchbase.Linq.Extensions
     {
         public static IQueryable<T> Queryable<T>(this IBucket bucket)
         {
-            return EntityFilterManager.ApplyFilters(new BucketQueryable<T>(bucket));
+            //TODO refactor so ClientConfiguration is injectable
+            return EntityFilterManager.ApplyFilters(new BucketQueryable<T>(bucket, new ClientConfiguration()));
         }
     }
 }
--- a/Src/Couchbase.Linq/BucketQueryExecutor.cs
+++ b/Src/Couchbase.Linq/BucketQueryExecutor.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using Common.Logging;
+using Couchbase.Configuration.Client;
 using Couchbase.Core;
 using Couchbase.Linq.Operators;
 using Couchbase.Linq.QueryGeneration;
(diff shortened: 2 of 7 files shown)


Src/Couchbase.Linq/Extensions/BucketExtensions.cs
```diff
@@ -8,7 +9,8 @@ public static class BucketExtensions
     {
         public static IQueryable<T> Queryable<T>(this IBucket bucket)
         {
-            return EntityFilterManager.ApplyFilters(new BucketQueryable<T>(bucket));
+            //TODO refactor so ClientConfiguration is injectable
+            return EntityFilterManager.ApplyFilters(new BucketQueryable<T>(bucket, new ClientConfiguration()));
```
brantburnett: @jeffrymorris,

I feel like this point here is where we should pull ClusterHelper configuration as the default, instead of a new configuration.  Otherwise we've lost support for pulling the ContractResolver from the current cluster configuration when using ClusterHelper.GetBucket("default").Queryable<T>().  It will always be the default CamelCasePropertyNamesContractResolver contract resolver.

Ideally, I think it would be great to have some way to get the contract resolver from IBucket.  Maybe have CouchbaseBucket implement an additional IBucketConfiguration interface that we could query for the configuration, something like that.  But that requires changes to the core client, which is obviously more difficult.

Brant

jeffrymorris: @brantburnett -

This class is now internal and then only place I see it being used was in the tests. The public interface here is now DBContext (Btw, shouldn't this be BucketContext?). I was going to add another param to Queryable so that it was injectible, but figured we could do that later...the LOE is the changes to 184 tests that depend on that method.

The goal is to make the ICluster implementation the dependency from  the provider, and remove the dependency on ClusterHelper - which can still be used at the application level. The ICluster object has a ClientConfiguration field which contains the necessary dependencies for the name resolver and translator. 

I agree about exposing the configuration info from IBucket, but it requires a change to couchbase-net-client, which would require a release before we could use it here. I think a better tactic is to do that in a later release.

-Jeff

brantburnett: I think BucketContext might make more sense than DbContext, and help avoid conflicts with EF in applications using both Couchbase and SQL/EF.

I certainly like removing the dependency on ClusterHelper from as much stuff as we can, especially within DbContext (or BucketContext) since it has a Configuration property.  And QueryFactory is internal.  But BucketExtensions is still a public API point.

If a user is getting an IBucket interface from ClusterHelper (instead of using DbContext), and then calls Queryable<T> on IBucket, I feel like their expectation will be that the serialization settings will be the one from the ClusterHelper.  But this implementation is currently not doing so in that use case.

Brant

jeffrymorris: I see, I think maybe the solution is to make BucketExtensions.Queryable internal (just for ease of testing), since the API we want public  is IDbContext. Doesn't make a lot of sense to have two different API's.

I agree BucketContext...I'll follow up with a commit that does this.

-Jeff

brantburnett: Sounds good to me.  It will just require refactoring a bunch of the BeerSampleTests, they all use BucketExtensions I think.

If we're renaming stuff, I wanted to float something else by you.  When I wrote the EntityFilterManager stuff, my head was in EF world.  But in hindsight, I wonder if naming that stuff DocumentXXX instead of EntityXXX might be more appropriate in Couchbase land.  Thoughts?

Brant

jeffrymorris: I concur!

brantburnett: Okay, I can handle that renaming, have it to you in a bit.

PR #55, opened 2015-10-10
Remove WhereMissing custom LINQ extension

### LINKED ISSUE #51: API Cleanup - Is WhereMissing Still Necessary?
Currently, there are two different LINQ methods to add a filter to only include documents missing a particular property:

1.  Use `.WhereMissing(p => p.Property)`
2.  Use N1Ql.IsMissing(p.Property) in one of two ways:
a.  `from p in bucketContext where N1Ql.IsMissing(p.Property) select p`
b.  `.Where(p => N1Ql.IsMissing(p.Property)`

Is having two methods of doing the same thing worthwhile?  It could add to long term maintenance issues as well as cause API confusion.

1.  .WhereMissing can only be used in lambda syntax, not in C# queries
2.  The inverse N1Ql.IsNotMissing is only available using syntax 2 above
3.  The similar methods IsValued and IsNotValued are also only available using syntax 2 above
4.  It will reduce the area required for long term support if removed

Thoughts?

brantburnett:
Fixes #51

Motivation
----------
The WhereMissing LINQ extension is supplanted by the N1Ql.IsMissing
function.  This function is more readable and supports usage in the select
projection and order by clauses.  It is also more consistent with
N1Ql.IsNotMissing, IsValued, and IsNotValued.

Modifications
-------------
Removed the WhereMissing extension, and associated classes and methods.
Updated the associated unit test to use N1Ql.IsMissing instead.


jeffrymorris:
Brant -

Thanks for the PR, this is now merged!

-Jeff

Issue #52, opened 2015-10-10
API Cleanup - Make QueryGeneration Namespace Internal


brantburnett:
Currently, several classes in the QueryGeneration namespace are public.  I think we should consider making them internal, so they are not part of the exposed API.

The biggest downside of this is the lack of extensibility. However:

1.  If we are going to allow extensions of the query generation process, I think some redesign is called for to provider cleaner extension points.  It will therefore be easier to do in the future if we don't need to worry about backwards compatibility.
2.  Extensions would also present a potential security concern.  One third party library could pump in a malicious extension which would then be unwittingly used by another third party library that was also using Linq2Couchbase.  So if we do provide the ability to add extensions, it should probably be at the BucketContext level for control.

brantburnett:
I also think that this applies to the Clauses and Operators namespaces as well

jeffrymorris:
I agree on both points! As for extension points I would rather expose public interfaces and/or "func"tories.

-Jeff

jeffrymorris:
This is merged, so closing this issue.

Thanks,

Jeff

PR #56, opened 2015-10-11
Make QueryGeneration, Clauses, and Operators internal

brantburnett:
Fixes issue #52

Motivation
----------
Reduce exposed API surface area.  Allow for the addition of more user
friendly extensibility points in the future.

Modifications
-------------
All public classes in the QueryGeneration, Clauses, and Operators
namespaces are now marked internal.  Added an InternalVisiblesTo attribute
for Moq to support tests that need to Mock any of these classes.


jeffrymorris:
Brant -

This is merged, thanks!

-Jeff

PR #60, opened 2015-10-15
LINQ-20: Add Save and Remove to IBucketContext

jeffrymorris:
Motivation
----------
Adds an additional WRITE API for users who do not want to do change
tracking (forthecoming).

Modifications
-------------
Added Save and Remove to IBucketContext and provided implementation in
BucketContext. Added a DocId attribute so the the key for the document Add
Save and Remove to IBucketContextcan be defined from one of the fields; I
considered making a string field called "_id" the default key-field, but
removed that logic - it may come back though. I split the
BucketContextTests into integrated and UT test classes. Finally, added a
couple of exceptions for handling specific use cases such as KeyNotFound
and when a key (id) field cannot be found. Note that Save and Remove throw
exceptions; so either it succeeded or it failed and the app needs to
handle the exception and failure case.

Result
------
You can now Save and Remove documents from the BucketContext.

### DIFF
--- a/Src/Couchbase.Linq.Tests/BucketContextTests.cs
+++ b/Src/Couchbase.Linq.Tests/BucketContextTests.cs
@@ -1,0 +1,102 @@
+﻿using Couchbase.Core;
+using Couchbase.IO;
+using Couchbase.Linq.Tests.Documents;
+using Moq;
+using NUnit.Framework;
+
+namespace Couchbase.Linq.Tests
+{
+    [TestFixture]
+    public class BucketContextTests
+    {
+        [Test]
+        public void GetDocumentId_When_Id_Field_DoesNotExist_Throw_MissingIndentifier_Exception()
+        {
+            var route = new Route();
+            var bucket = new Mock<IBucket>();
+            var ctx = new BucketContext(bucket.Object);
+            Assert.Throws<DocumentIdMissingException>(()=>ctx.GetDocumentId(route));
+        }
+
+        [Test]
+        public void GetDocumentId_When_DocId_Exists_Use_It()
+        {
+            var beer = new Beer();
+            var bucket = new Mock<IBucket>();
+            var ctx = new BucketContext(bucket.Object);
+            var id = ctx.GetDocumentId(beer);
+            Assert.AreEqual("name", id);
+        }
+
+       [Test]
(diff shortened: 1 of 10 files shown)


Src/Couchbase.Linq.Tests/App.config
```diff
@@ -14,7 +14,7 @@
   <couchbase enableConfigHeartBeat="false">
       <servers>
         <!-- changes in appSettings section should also be reflected here -->
-        <add uri="http://localhost:8091"></add>
+        <add uri="http://192.168.77.101:8091"></add>
```
brantburnett: Jeff,

Did you mean to leave this IP in the .config file?

Brant

jeffrymorris: Nope!

On Oct 15, 2015 8:24 AM, "Brant Burnett" <notifications@github.com> wrote:
>
> In Src/Couchbase.Linq.Tests/App.config:
>
> > @@ -14,7 +14,7 @@
> >    <couchbase enableConfigHeartBeat="false">
> >        <servers>
> >

Src/Couchbase.Linq/Mapping/DocIdAttribute.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+
+namespace Couchbase.Linq.Mapping
+{
+    /// <summary>
+    /// Provides a means of mapping a field to a Couchbase document key or Id.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property)]
+    public class DocIdAttribute : Attribute
```
brantburnett: Jeff,

I think you may have accidentally left this old file in the pull, unless you were planning on using it for something else.

Brant

brantburnett:
Jeff,

Overall I think this looks really good.  I just have a couple of thoughts.  But they're more implementation than API related, so they could easily be held for a future version.

1.  Some kind of caching mechanism for the identified document ID attribute would improve performance, so it doesn't need to scan with reflection on every save.

2.  The only member name format currently supported is JsonProperty, not DataMember.  We should probably look at supporting the same member name resolution system as the LINQ side.  Though admittedly that is somewhat limited right now, too.

Brant

Issue #68, opened 2015-10-21
META() does not serialize correctly


jeffrymorris:
~~https://issues.couchbase.com/browse/LINQ-23~~

https://issues.couchbase.com/browse/NCBC-1021

brantburnett:
@jeffrymorris 

The problem is probably the way we're wrapping it.  If you put it inside a new like this I believe it works:

`new with {.meta = N1Ql.Meta(p)}`

We probably need to add handling for this use case.

Brant

jeffrymorris:
@brantburnett  -

It appears to be something on the SDK side; the generated N1QL statement is valid and works in CBQ:

cbq> SELECT META(`Extent1`) FROM `beer-sample` as `Extent1` LIMIT 1;
{
    "requestID": "d3fe60fc-6742-49e3-8cf6-110669a040ef",
    "signature": {
        "$1": "object"
    },
    "results": [
        {
            "$1": {
                "cas": 1.4454516563845448e+18,
                "flags": 0,
                "id": "21st_amendment_brewery_cafe-563_stout",
                "type": "json"
            }
        }
    ],
    "status": "success",
    "metrics": {
        "elapsedTime": "52.771019ms",
        "executionTime": "52.730494ms",
        "resultCount": 1,
        "resultSize": 214
    }
}

I'll dig through the SDK and see why it's not mapping correctly (the "$1" looks suspect, though).

-Jeff

brantburnett:
So, after some further research I think this really is an issue we need to handle on the Linq2Couchbase library side.  It isn't just affecting the META keyword, but basically other use cases that don't include a select project.

This use case is working:

from beer in context.Query<Beer>()
select beer;

It works because we automatically add ".*" in the query:

SELECT `Extent1`.* FROM `beer-sample` as `Extent1`

This projects all the properties of the beer out into the result object for us, so we don't have to do it manually.

Additionally, with a small tweak we can apply the same approach to the META function by adding ".*".

SELECT META(`Extent1`).* FROM `beer-sample` as `Extent2`

This will fix the META keyword if we are attempting to return it directly, with "new {...}" in the select projection.  However, there are other use cases that this doesn't fix.

For example, trying to select a single property directly from the object:

from b in context.Query<Beer>()
select b.Abv;

[...]

PR #71, opened 2015-10-24
Issue #68: Fix handling for queries without a new object select projection

brantburnett:
Motivation
----------
Queries that don't include a new object in the select projection (i.e.
`select new {...}`) are failing to deserialize.  The result is returned by
N1QL as an array of objects with a $1 property, and the expected value is
stored in this property.  However, LINQ deserialization is expecting an
array of the plain value, without the intermediate wrapper object.

Modifications
-------------
Added a property to N1QlQueryModelVisitor, ResultExtractionRequired, which
indicates to the BucketQueryExecutor that a property extraction is
required during deserialization.  This property is always named "result".
BucketQueryExecutor now uses this property to recognize that it needs to
deserialize an array of SimpleResult<T> instead of any array of T, then
extracts the result property.

Repurposed "ArrayPropertyExtractionPart", which was formerly used to
perform a similar function when dealing with array subqueries and Any/All.
It is now "PropertyExtractionPart", and is used for array subqueries as
well as main queries where property extraction is required.

Previously, aggregates were manually having property extraction performed
on them.  They now use the new property extraction approach, which
simplifies the system and should make things less confusing going forward.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
@@ -188,6 +188,25 @@ namespace Couchbase.Linq.IntegrationTests
         }
 
         [Test]
+        public void Map2PocoTests_Explain_QueryWithPropertyExtraction()
+        {
+            using (var cluster = new Cluster(TestConfigurations.DefaultConfig()))
+            {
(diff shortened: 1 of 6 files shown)


jeffrymorris:
Brant -

This is merged, thanks!

-Jeff

PR #73, opened 2015-10-26
Fixed array subqueries using select expressions without new object

brantburnett:
Motivation
------------
Array subqueries with simple select expressions are incorrectly including
"as `result`" in the query.

ex.
```
from brewery in QueryFactory.Queryable<Brewery>(mockBucket.Object)
select new { name = brewery.Name, addresses = brewery.Address.Select(p => "Address " + p) };
```

Modifications
---------------
Exclude array subqueries from the logic adding a PropertyExtractionPart to
the QueryPartsAggregator.  Added a unit test for this scenario.


jeffrymorris:
@brantburnett 

This is merged.

Thanks,

Jeff

PR #75, opened 2015-10-26
Completed more documentation, plus some formatting cleanup

brantburnett:
Modifications
-------------
Completed documenation for:
- Array filtering, projections, and sorting
- Date handling
- String handling
- Math functions
- Sorting and paging
- Grouping and aggregates

Updated Document Filters documentation to reference creating custom
attributes inherited from DocumentFilterAttribute.  Also removed the
assertion that JsonProperty attribute is not supported after confirming
via testing that it is supported.

Miscellaneous formatting improvements.


jeffrymorris:
Thanks Brant!

Issue #76, opened 2015-10-26
String Concatenation With Non-String Values


brantburnett:
String concatenation doesn't currently work with non-string values/properties.

If you concatenate with +, the generated N1QL is using + insted of ||.

If you attempt to use .ToString(), then there is an exception generating the query.

This could be a significant use case, particularly for compound keys.  For example, if you build your keys with a combination of the type and GUID columns, but store just the GUID in related documents.  When doing a JOIN or NEST operation, you will need to concatenate the a type string constant with the GUID to build the key of the document being joined.

There is an example of this in the docs, https://github.com/couchbaselabs/Linq2Couchbase/blob/master/docs/joins.md, which is currently not working.

PR #78, opened 2015-10-26
Issue #76: Fixed string concatenation for non-strings and ToString

brantburnett:
Motivation
----------
When building compound keys for joins and nests, a common use case may be
to combine strings with non-string properties such as numbers and GUIDs.

Modifications
-------------
Changed handling of binary + expression to use string concatenation
operator if either operand is a string.  Previously it was only doing this
if both operands were strings.

Added a method call translator for .ToString() that uses the N1QL
TOSTRING() function.  This required making a tweak to the
DefaultMethodCallTranslatorProvider to check for method call translators
against virtual base methods.  This allowed registering the method call
translator only on Object.ToString while applying to all inherited
types.

Results
-------
String handling for non-string types now behaves in the way a user would
expect.


jeffrymorris:
@brantburnett -

This is merged, thanks!

-Jeff

PR #81, opened 2015-11-14
Support IExtendedTypeSerializer member name resolution

brantburnett:
Motivation
----------
The current implementation for resolving the names of members (such as
properties) only supports Newtonsoft.Json based contract resolvers.  It
uses obsolete settings from the SDK ClientConfiguration.  This basically
limits support to clients that don't use a custom ITypeSerializer
implementation.

Modifications
-------------
If ClientConfiguration.Serializer returns an IExtendedTypeSerializer, use
the provided GetMemberName method to resolve member names.  This is done
through ExtendedTypeSerializerMemberNameResolver, which maps GetMemberName
to the IMemberNameResolver interface.

If ClientConfiguration.Serializer doesn't return an
IExtendedTypeSerializer, fallback to the former implementation of
JsonNetMemberNameResolver.

Results
-------
Newer SDK users who implement IExtendedTypeSerializer for custom
serialization will now be fully compatible with Linq2Couchbase.

SDK users who were formerly using the SDK DefaultSerializer implementation
will use the new serializer.  However, the SDK implements the same logic
as JsonNetMemberNameResolver, so the change is transparent.

SDK users who implement an older ITypeSerializer will also see the
behavior unchanged until they update to an IExtendedTypeSerializer.

### DIFF
--- a/Src/Couchbase.Linq.UnitTests/BucketQueryExecutorEmulator.cs
+++ b/Src/Couchbase.Linq.UnitTests/BucketQueryExecutorEmulator.cs
@@ -60,7 +60,7 @@ namespace Couchbase.Linq.UnitTests
         {
             var queryGenerationContext = new N1QlQueryGenerationContext()
             {
-                MemberNameResolver = new JsonNetMemberNameResolver(Test.ContractResolver),
+                MemberNameResolver = Test.MemberNameResolver,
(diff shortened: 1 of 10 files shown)


jeffrymorris:
Brant -

This is merged!

Thanks,

jeff

Issue #85, opened 2015-12-10
Projection to strong types


wiciul:
If it would be possible it would be great to have possibility to project to strong and not only anonymous types.

brantburnett:
@wiciul 

Can you send some example code that is not working.  It should function as long as you are using the default constructor (no parameters) and then assigning values in a { ... } block.

Brant

wiciul:
var logEntry = _unresolvedCoverageLogContext
                .Query<UnresolvedCoverageLogEntry>()
                .Where(e => e.Type == "Reuters.Tornado.unresolvedcoverage.addUnresolvedCoverageRequest")
                .Select(e => new AddUnresolvedCoverageRequest
                {
                    Date = e.Date
                }).ToList();


Type : System.NotSupportedException, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089
Message : The expression 'new AddUnresolvedCoverageRequest() {Date = [e].Date}' (type: System.Linq.Expressions.MemberInitExpression) is not supported by this LINQ provider.
Source : Remotion.Linq
Help link : 
Data : System.Collections.ListDictionaryInternal
TargetSite : TResult VisitUnhandledItem[TItem,TResult](TItem, System.String, System.Func`2[TItem,TResult])
HResult : -2146233067
Stack Trace :    at Remotion.Linq.Parsing.ThrowingExpressionTreeVisitor.VisitUnhandledItem[TItem,TResult](TItem unhandledItem, String visitMethod, Func`2 baseBehavior)
   at Remotion.Linq.Parsing.ThrowingExpressionTreeVisitor.VisitMemberInitExpression(MemberInitExpression expression)
   at Remotion.Linq.Parsing.ExpressionTreeVisitor.VisitExpression(Expression expression)
   at Couchbase.Linq.QueryGeneration.N1QlExpressionTreeVisitor.VisitExpression(Expression expression)
   at Couchbase.Linq.QueryGeneration.N1QlExpressionTreeVisitor.GetN1QlExpression(Expression expression, N1QlQueryGenerationContext queryGenerationContext)
   at Couchbase.Linq.QueryGeneration.N1QlQueryModelVisitor.GetN1QlExpression(Expression expression)
   at Couchbase.Linq.QueryGeneration.N1QlQueryModelVisitor.GetSelectParameters(SelectClause selectClause, QueryModel queryModel)
[...]

PR #86, opened 2015-12-11
Issue #85: Support for strongly typed select projections

brantburnett:
Motivation
----------
In some circumstances, it is desirable to project your documents onto
strongly typed POCOs which don't match your document structure.
Currently, the only way to select directly onto strongly typed POCOs is if
they are your documents.

Modifications
-------------
Added limited support for the MemberInitExpression in
N1QlExpressionTreeVisitor.  It is supported for both creating new objects
in the query ({"x": y} syntax) and in the main select clause (y as x
syntax).  This means it is also supported in array-based subqueries.

This implementation doesn't support constructor parameters, the
constructor must be parameterless.  Also, it doesn't support other forms
of MemberInitExpression such as list initialization.

Implemented appropriate unit tests.

Results
-------
Strongly typed POCOs can now be used in select projections.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
@@ -61,6 +61,26 @@ namespace Couchbase.Linq.IntegrationTests
         }
 
         [Test]
+        public void Map2PocoTests_StronglyTyped_Projections()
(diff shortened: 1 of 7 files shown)


brantburnett:
Closing this pull request so we can pull the work again against a different branch.

Issue #89, opened 2015-12-26
Couchbase Server 4.1 Error Executing Query With Prefiltered All Subquery


brantburnett:
The integration test AnyAllTests_AllNestedArrayPrefiltered is failing.  The generated N1QL query is:

```
SELECT `Extent1`.`name` as `name`, `Extent1`.`address` as `address`
FROM `beer-sample` as `Extent1`
WHERE ((`Extent1`.`type` = 'brewery')
AND EVERY `Extent2` IN (ARRAY `Extent2` FOR `Extent2` IN `Extent1`.`address` WHEN (`Extent2` = '563 Second Street') END) SATISFIES (`Extent2` = '101 Fake Street') END)
ORDER BY `Extent1`.`name` ASC
```

This test was not failing against Couchbase Server 4.0.  The reuse of Extent2 outside the ARRAY subquery for the EVERY statement is apparently not allowed by the the Couchbase Server 4.1 query engine.

jeffrymorris:
@brantburnett -

I opened a server ticket for an explanation on this: https://issues.couchbase.com/browse/MB-17170

-Jeff

jeffrymorris:
@brantburnett -

It looks like EVERY and ANY should introduce a new variable "Extent3" in the query:

{code}
SELECT `Extent1`.`name` as `name`, `Extent1`.`address` as `address`
 FROM `beer-sample` as `Extent1`
 WHERE ((`Extent1`.`type` = 'brewery')
 AND EVERY `Extent3` IN (ARRAY `Extent2` FOR `Extent2` IN `Extent1`.`address` WHEN (`Extent2` = '563 Second Street') END) SATISFIES (`Extent3` = '101 Fake Street') END)
 ORDER BY `Extent1`.`name` ASC
{code}

That's the official word from the query team.

-Jeff

brantburnett:
@jeffrymorris 

Yeah, I thought they would say that.  I've just played with it a little so far, but it looks like now you can reference identifiers from outside the ARRAY statement inside the ARRAY statement.  Thus they can't allow identifier conflicts anymore.  It's going to be tricky to fix in LINQ, though.

Brant

brantburnett:
Fixed, see 48a80fa

Issue #43, opened 2015-10-03
Relinq 2.0 compatibility issues


jeffrymorris:
See here: https://issues.couchbase.com/browse/LINQ-15

jeffrymorris:
https://github.com/couchbaselabs/Linq2Couchbase/commit/a59b14c72638ec4aedaf23729b4e7be3b1e330c7

PR #90, opened 2015-12-26
Issue #43, LINQ-15: Upgraded to Relinq 2.0

brantburnett:
Motivation
----------
Relinq 2.0 is not backwards compatible with Relinq 1.0.  Other frameworks,
such as Entity Framework 7, are using Relinq 2.0.  In order to be
compatible for side-by-side use, we must also use Relinq 2.0.

Modifications
-------------
Updated to Nuget to Relinq 2.0.1.  Made lots of adjustments to consume the
new Relinq API.  Most of the changes were changes to names, such as all of
the ExpressionTreeVisitors becoming ExpressionVisitors.  There were also
some method signature changes.

Te other significant change was testing or special Relinq Expressions.
They used to have special expression type values.  Now they are all
Extension expression types, and we need to detect using the .Net type.

Results
-------
System now uses Relinq 2.0, and should be compatible with EntityFramework
7 and other libraries using Relinq 2.0.  All tests pass, except for the
one in outstanding issue #89, which was failing before the upgrade.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/Couchbase.Linq.IntegrationTests.csproj
+++ b/Src/Couchbase.Linq.IntegrationTests/Couchbase.Linq.IntegrationTests.csproj
@@ -81,9 +81,9 @@
       <HintPath>..\packages\NUnit.2.6.4\lib\nunit.framework.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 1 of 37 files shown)


jeffrymorris:
@brantburnett -

This is merged, can you please close the PR?

Thanks!

-Jeff

PR #92, opened 2015-12-26
Issue #85: Support for strongly typed select projections

brantburnett:
Motivation
----------
In some circumstances, it is desirable to project your documents onto
strongly typed POCOs which don't match your document structure.
Currently, the only way to select directly onto strongly typed POCOs is if
they are your documents.

Modifications
-------------
Added limited support for the MemberInitExpression in
N1QlExpressionTreeVisitor.  It is supported for both creating new objects
in the query ({"x": y} syntax) and in the main select clause (y as x
syntax).  This means it is also supported in array-based subqueries.

This implementation doesn't support constructor parameters, the
constructor must be parameterless.  Also, it doesn't support other forms
of MemberInitExpression such as list initialization.

Implemented appropriate unit tests.

Results
-------
Strongly typed POCOs can now be used in select projections.

### DIFF
--- a/Src/Couchbase.Linq/QueryGeneration/N1QLQueryModelVisitor.cs
+++ b/Src/Couchbase.Linq/QueryGeneration/N1QLQueryModelVisitor.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using Couchbase.Annotations;
 using Couchbase.Core.Serialization;
 using Couchbase.Linq.Clauses;
 using Couchbase.Linq.Operators;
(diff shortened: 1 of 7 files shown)


brantburnett:
This is my previous work, but now refactored for Relinq 2.0

Thanks,
Brant

Src/Couchbase.Linq/QueryGeneration/N1QLQueryModelVisitor.cs
```diff
@@ -829,55 +830,55 @@ private N1QlFromQueryPart ParseNestClause(NestClause nestClause)
...
-                {
-                    ItemName = GetExtentName(nestClause),
-                    Value =
-                        string.Format("ARRAY {0} FOR {0} IN {1} WHEN {2} END",
-                            GetExtentName(subQuery.QueryModel.MainFromClause),
-                            genItemName,
-                            whereClauseString)
-                };
+                    var letPart = new N1QlLetQueryPart()
+                    {
```
jeffrymorris: It looks like there is extra leading indentation in several places on this file. Is this on purpose?

brantburnett: @jeffrymorris 

Good catch, it was an artifact of the rebase to include the Relinq 2.0 work.  I've corrected it.

Thanks,
Brant

jeffrymorris:
@brantburnett -

Thanks!

-Jeff

Issue #88, opened 2015-12-23
Enums Not Supported


jawang35:
I couldn't find information in the documentation about this anywhere. Are there any plans to offer the ability to filter enum properties and enum flags?

brantburnett:
@jawang35 

You raise a very good question.  It's certainly an excellent feature I'd like to see added.  The trick is going to be matching the way enumerations are represented in N1QL with the way they're being serialized in the documents.

Can you provide some examples of the serialization method you're using in your use case?  For example, are you using StringEnumConverter in Json.net?

jeffrymorris:
@jawang35, @brantburnett -

Here is a blog regarding enum support in EF: http://blogs.msdn.com/b/efdesign/archive/2011/06/29/enumeration-support-in-entity-framework.aspx

Checkout the "Query" and "HasFlags" support sections specifically, is this what you envision the support would cover? HasFlags would have to be translated into multiple "OR"s in the predicate.

-Jeff

jawang35:
@brantburnett , @jeffrymorris 

I am using StringEnumConverter and I did imagine it would be implemented just like enums are supported in EF. The only difference being possibly allowing the enums to be stored as strings instead of their underlying value type when StringEnumConverter is used. If that is the case then translating the predicate will have to take into consideration whether the converter is specified. I think by default (when StringEnumConverter is not used) the underlying value type will have to be stored because of the ability for two enum members to have the same underlying value.

brantburnett:
@jawang35 

I've put in pull request #96, which should add the support you are looking for to the next release.  It will allow your use of StringEnumConverter to pass through to the N1QL query, so long as you are using == or != comparisons.

Regarding enum flags, that's a bit more complicated.  The work that was done should support == and != checking of flags, so long as the JSON converter supports them, and it always writes the flags consistently.  Note that I don't think StringEnumConverter supports flags.

However, it still won't support doing bitwise operations on the flags.  At this time, I can't find any documentation where N1QL supports bitwise operations in the query.  My recommendation at this time, unfortunately, is to do any bitwise filtering in code rather than in query:

```
var docs = context.Query<Document>()
    .Where(p => p.Age > 10)
    .AsEnumerable()
    .Where(p => (p.Flags & Flag.One) == Flag.One);
```

Note that any lambdas applied after an AsEnumerable() call are executed in code, after the query is complete.  This is less performant than filtering done in the query, but required to do things unsupported by the LINQ query engine.

Thanks,
Brant

PR #96, opened 2015-12-29
Issue #88: Support for string enumeration comparisons

brantburnett:
Motivation
----------
A common use case for enumerations is to store them in the JSON document
as strings rather than integers.  With Json.Net, this is usually done with
StringEnumConverter.  In order to filter N1QL documents based on these
enumerations in WHERE clauses, the outputed N1QL query must compare using
the string representation instead of the integer representation.

Modifications
-------------
Whenever a == or != comparison is performed against an enumeration
constant, convert the expression in EnumComparisonExpressionTransformer
to use the enumeration type instead of the base integer type.  This allows
the constant to be recognized as an enumeration constant later by
N1QlExpressionTreeVisitor.

When an enumeration constant is encountered by N1QlExpressionTreeVisitor,
use the configured ITypeSerializer to determine how to output the value to
N1Ql.  This is done by first serializing to JSON, then deserializing back
to a .Net type.  This gets a string or integer value, depending on how the
type serializer is configured.  This can then be output to the N1QL query
like any other constant.

Results
-------
Equal and not equal comparisons of enumeration values are now supported
when using a StringEnumConverter.  This should theoretically extend to any
other non-standard JSON conversion, so long as the non-standard
serialization is applied directly at the enumeration type.  It would not
work, for example, if a custom serializer was applied to the document
itself that changed how the enumeration property is serialized.

Note that greater than and less than comparisons are not supported with
StringEnumConverter.  They are still supported when working with numeric
serialization.  However, they simply don't translate to N1QL from .Net as
a string, because the .Net comparison greater than or less than comparison
would still be based on the numeric value.  In N1QL, it would be based on
the string value from the serialization layer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ NuGet will install the package and all dependencies. Once you have the resolved 
 - [Any and All](https://github.com/couchbaselabs/Linq2Couchbase/blob/master/docs/any-all.md)
 - [Testing For NULL And MISSING Attributes](https://github.com/couchbaselabs/Linq2Couchbase/blob/master/docs/null-missing-valued.md)
 - [The META Keyword](https://github.com/couchbaselabs/Linq2Couchbase/blob/master/docs/meta-keyword.md)
+- [Working With Enumerations](https://github.com/couchbaselabs/Linq2Couchbase/blob/master/docs/enum.md)
 
 ##Building From Source##
 
--- a/Src/Couchbase.Linq.IntegrationTests/Couchbase.Linq.IntegrationTests.csproj
+++ b/Src/Couchbase.Linq.IntegrationTests/Couchbase.Linq.IntegrationTests.csproj
@@ -99,6 +99,8 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="BeerSample.cs" />
+    <Compile Include="Documents\BeerStyle.cs" />
+    <Compile Include="Documents\BeerWithEnum.cs" />
     <Compile Include="QueryTests.cs" />
     <Compile Include="BucketContextTests.cs" />
     <Compile Include="Documents\BeerFiltered.cs" />
--- a/Src/Couchbase.Linq.IntegrationTests/Documents/BeerStyle.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/Documents/BeerStyle.cs
@@ -1,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace Couchbase.Linq.IntegrationTests.Documents
+{
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum BeerStyle
+    {
+        Porter,
+
+        [EnumMember(Value="Oatmeal Stout")]
+        OatmealStout,
+
+        [EnumMember(Value="Pumpkin Beer")]
+        PumpkinBeer
+    }
+}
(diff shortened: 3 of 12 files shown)


jeffrymorris:
@brantburnett -

A couple of tests are failing on my machine:
- Test_NoEnumConverterOnNullable_WhereEqualToNull
- Test_NoEnumConverterOnNullable_WhereNotEqualToNull
- Test_StringEnumConverterOnNullable_WhereEqualToNull
- Test_StringEnumConverterOnNullable_WhereNotEqualToNull

All fail with the same exception:
```
System.ArgumentNullException : Value cannot be null.
Parameter name: value

   at System.RuntimeType.GetEnumName(Object value)
   at Couchbase.Linq.QueryGeneration.ExpressionTransformers.EnumComparisonExpressionTransformer.MakeEnumComparisonExpression(Expression operand, Object enumValue, ExpressionType comparisonType) in c:\Users\jmorris\repos\Linq2Couchbase\Src\Couchbase.Linq\QueryGeneration\ExpressionTransformers\EnumComparisonExpressionTransformer.cs:line 87
   at Couchbase.Linq.QueryGeneration.ExpressionTransformers.EnumComparisonExpressionTransformer.Transform(BinaryExpression expression) in c:\Users\jmorris\repos\Linq2Couchbase\Src\Couchbase.Linq\QueryGeneration\ExpressionTransformers\EnumComparisonExpressionTransformer.cs:line 35
   at Remotion.Linq.Parsing.ExpressionVisitors.TransformingExpressionVisitor.Visit(Expression expression)
   at System.Linq.Expressions.ExpressionVisitor.VisitLambda[T](Expression`1 node)
   at Remotion.Linq.Parsing.ExpressionVisitors.TransformingExpressionVisitor.Visit(Expression expression)
   at System.Linq.Expressions.ExpressionVisitor.VisitUnary(UnaryExpression node)
   at Remotion.Linq.Parsing.ExpressionVisitors.TransformingExpressionVisitor.Visit(Expression expression)
   at System.Linq.Expressions.ExpressionVisitor.VisitArguments(IArgumentProvider nodes)
   at System.Linq.Expressions.ExpressionVisitor.VisitMethodCall(MethodCallExpression node)
   at Remotion.Linq.Parsing.ExpressionVisitors.TransformingExpressionVisitor.Visit(Expression expression)
   at System.Linq.Enumerable.Aggregate[TSource,TAccumulate](IEnumerable`1 source, TAccumulate seed, Func`3 func)
   at Remotion.Linq.Parsing.Structure.ExpressionTreeParser.ParseTree(Expression expressionTree)
   at Remotion.Linq.Parsing.Structure.QueryParser.GetParsedQuery(Expression expressionTreeRoot)
   at Couchbase.Linq.UnitTests.N1QLTestBase.CreateN1QlQuery(IBucket bucket, Expression expression) in c:\Users\jmorris\repos\Linq2Couchbase\Src\Couchbase.Linq.UnitTests\N1QLTestBase.cs:line 39
   at Couchbase.Linq.UnitTests.QueryGeneration.EnumTests.Test_NoEnumConverterOnNullable_WhereEqualToNull() in c:\Users\jmorris\repos\Linq2Couchbase\Src\Couchbase.Linq.UnitTests\QueryGeneration\EnumTests.cs:line 115
```

Debugging through the enum value is null here: https://github.com/couchbaselabs/Linq2Couchbase/pull/96/files#diff-3c274e1ac7db21aa8f19baafb2496db4R87

Thanks,

Jeff

brantburnett:
@jeffrymorris 

Interesting.  I actually thought that those tests would fail with this exact error when I wrote the tests.  But then they didn't on my machine, so I thought maybe LINQ was handling it for me.  I wonder if it's a difference between the old C# compiler and Roslyn, since I'm using VS2015.  I'll give it a look tonight.

Thanks,
Brant

brantburnett:
@jeffrymorris 

I've confirmed that the difference was between compiling the tests in VS2013 versus 2015.  I've fixed the bug, and confirmed that it works correctly in both version of VS.

Thanks,
Brant

jeffrymorris:
@brantburnett -

This is merged, thanks for the submission!

-Jeff

PR #97, opened 2015-12-31
Refactor string comparison expression conversion

brantburnett:
Motivation
----------
String comparisons using String.Compare and String.CompareTo are being
converted to StringComparisonExpression objects.  However, it is being
done inside N1QlExpressionTreeVisitor.  The more maintainable approach is
to use an ExpressionTransformer that is used earlier in the query parsing
process.

Modifications
-------------
Moved logic that converts string comparisons from
N1QlExpressionTreeVisitor to StringComparisonExpressionTransformer.

Updated the implementation of StringComparisonExpression to be a proper
Extension expression.

Moved string comparison unit tests to a new class to specifically test the
expression transformer.

Results
-------
String comparison conversions are more maintainable.  All tests are still
passing.


jeffrymorris:
@brantburnett -

This is merged, thanks!

-Jeff

PR #99, opened 2016-01-02
Support for N1QL date functions

brantburnett:
Motivation
----------
N1QL supports several date functions, which don't necessarily translate
directly to .Net date functions.  Supporting them will help users by
allowing filters to be applied during a query, rather than after data is
delivered to the application.

Modifications
-------------
Designed a new system for defining N1QL functions in .Net by decorating
static methods with N1QlFunctionAttribute.  It is then assumed that there
is a one-to-one mapping between the .Net parameters and the parameters of
the N1QL function.

Added additional methods to N1QlFunctions to act as the equivalent to
DATE_DIFF_STR, DATE_ADD_STR, DATE_PART_STR, and DATE_TRUNC_STR.  This
includes a new enumeration, N1QlDatePart, for the date part parameter.
The string based N1QL functions were chosen instead of the Unix
milliseconds version because they can be more easily parsed by Json.Net.

To help with code organization, split N1QlFunctions into different files
using partial classes.

Results
-------
N1QL functions for date/time addition, differences, part extraction, and
truncation are now supported.  There is a new design method which can be
used to easily add support for more N1QL functions like these in the
future.

### DIFF
--- a/Src/Couchbase.Linq/N1QlDatePart.cs
+++ b/Src/Couchbase.Linq/N1QlDatePart.cs
@@ -1,0 +1,82 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace Couchbase.Linq
+{
+    /// <summary>
+    /// Represents date parts for calls to date related <see cref="N1QlFunctions"/>.
+    /// </summary>
+    /// <remarks>
+    /// Different date related functions are compatible with different date parts.
+    /// For details, see the N1QL documentation.
+    /// </remarks>
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum N1QlDatePart
+    {
+        [EnumMember(Value = "millennium")]
+        Millennium,
+
+        [EnumMember(Value = "century")]
+        Century,
+
+        [EnumMember(Value = "decade")]
+        Decade,
+
+        [EnumMember(Value = "year")]
+        Year,
+
+        [EnumMember(Value = "quarter")]
+        Quarter,
+
+        [EnumMember(Value = "month")]
+        Month,
+
+        [EnumMember(Value = "week")]
+        Week,
+
+        [EnumMember(Value = "day")]
+        Day,
+
+        [EnumMember(Value = "hour")]
+        Hour,
+
+        [EnumMember(Value = "minute")]
+        Minute,
+
+        [EnumMember(Value = "second")]
+        Second,
+
+        [EnumMember(Value = "millisecond")]
+        Millisecond,
+
+        [EnumMember(Value = "doy")]
+        DayOfYear,
+
+        [EnumMember(Value = "dow")]
+        DayOfWeek,
+
+        [EnumMember(Value = "iso_week")]
+        IsoWeek,
+
+        [EnumMember(Value = "iso_year")]
+        IsoYear,
+
+        [EnumMember(Value = "iso_dow")]
+        IsoDayOfWeek,
+
+        [EnumMember(Value = "timezone")]
+        TimeZone,
+
+        [EnumMember(Value = "timezone_hour")]
+        TimeZoneOffsetHour,
+
+        [EnumMember(Value = "timezone_minute")]
+        TimeZoneOffsetMinute
+    }
+}
--- a/Src/Couchbase.Linq/N1QlFunctionAttribute.cs
+++ b/Src/Couchbase.Linq/N1QlFunctionAttribute.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Couchbase.Linq
+{
+    /// <summary>
+    /// Decorates a method that is converted to a N1QL function call.  Each parameter in .Net
+    /// is expected to correspond directly to a parameter in N1QL, in the same order.
+    /// </summary>
+    /// <remarks>
+    /// Should only be used to decorate static methods.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Method)]
+    internal class N1QlFunctionAttribute : Attribute
+    {
+        /// <summary>
+        /// Name of the function in N1QL.
+        /// </summary>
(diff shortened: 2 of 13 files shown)


Src/Couchbase.Linq/QueryGeneration/DefaultMethodCallTranslatorProvider.cs
```diff
@@ -48,48 +52,61 @@ public IMethodCallTranslator GetTranslator(MethodCallExpression methodCallExpres
...
-            IMethodCallTranslator translator;
-            if (Registry.TryGetValue(key, out translator))
-            {
-                return translator;
-            }
+                IMethodCallTranslator translator;
+                if (Registry.TryGetValue(key, out translator))
+                {
+                    return translator;
+                }
```
jeffrymorris: Looks like there is an extra \t here?

Src/Couchbase.Linq/QueryGeneration/DefaultMethodCallTranslatorProvider.cs
```diff
@@ -48,48 +52,61 @@ public IMethodCallTranslator GetTranslator(MethodCallExpression methodCallExpres
...
-            if (translator != null)
-            {
-                return translator;
-            }
+                // Check any interfaces that may have a matching method
+                translator = GetItemFromInterfaces(key);
+                if (translator != null)
+                {
+                    return translator;
+                }
```
jeffrymorris: Extra \t?

brantburnett: @jeffrymorris 

No extra \t, the entire method was indented to add the lock (Registry) wrapper.  I added this because we were modifying the registry, not just reading from it anymore.  Probably should have had it before, though.

Brant

Src/Couchbase.Linq/N1QlFunctionAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    /// <summary>
+    /// Decorates a method that is converted to a N1QL function call.  Each parameter in .Net
+    /// is expected to correspond directly to a parameter in N1QL, in the same order.
+    /// </summary>
+    /// <remarks>
+    /// Should only be used to decorate static methods.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Method)]
+    internal class N1QlFunctionAttribute : Attribute
```
jeffrymorris: Nice!

Src/Couchbase.Linq/N1QlFunctions.Core.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// Implements static helper methods for N1QL queries
+    /// </summary>
+    public static partial class N1QlFunctions
+    {
+        /// <summary>
+        /// Shortcut for creating an error for methods that are only supported in N1QL, not in .Net.
+        /// </summary>
+        private static Exception NotImplementedError()
+        {
+            return new NotImplementedException("This method may only be used for generating N1QL queries.");
```
jeffrymorris: I am not sure I understand the message here...not implemented means, well, not implemented. How can the method this exception is thrown from have an implementation (assumed from "This method may only be used for generating N1QL queries.")?

brantburnett: @jeffrymorris 

Maybe we need to make it clearer.  Any calls to it are converted to N1QL by the query parsing process.  But you can't call it directly in .Net, only use it in lambda expressions that are never called.

Looking at the way EF does it, maybe it should be a NotSupportedException instead?

http://entityframework.codeplex.com/SourceControl/latest#src/EntityFramework.SqlServer/SqlFunctions.cs

Brant

Src/Couchbase.Linq/N1QlDatePart.cs
```diff
@@ -0,0 +1,82 @@
...
+        Minute,
+
+        [EnumMember(Value = "second")]
+        Second,
+
+        [EnumMember(Value = "millisecond")]
+        Millisecond,
+
+        [EnumMember(Value = "doy")]
+        DayOfYear,
```
jeffrymorris: "doy" and "dow", oh mai :)

Src/Couchbase.Linq/N1QlFunctions.Core.cs
jeffrymorris: Ah, I think I got it. Yeah, I think not supported would make more sense. 

-Jeff

brantburnett: @jeffrymorris

Pushed up changes to make it a NotSupportedException.

Brant

Src/Couchbase.Linq/N1QlFunctions.DateTime.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Couchbase.Linq
+{
+    public partial class N1QlFunctions
+    {
```
jeffrymorris: I have never been a huge fan because the spread the code out across multiple files in the directory making it hard to follow the code (where is this method coming from?). That being said, I am always willing to change my way with a strong argument ;)

Src/Couchbase.Linq/QueryGeneration/DefaultMethodCallTranslatorProvider.cs
jeffrymorris: Gotcha!

brantburnett:
@jeffrymorris 

Generally speaking, I'm of the same opinion.  However, I've also found that when files start to get too large it can also be detrimental.  Resharper starts to run slowly.  If multiple people work in the same file (or one person on the file in different version branches) then merge conflicts are more common.  Finding the code you're looking for through all the region directives gets difficult.

So when I see a file getting really big, and if it has a very logical and highly organized way to break it up, I sometimes go for it.  The N1QlFunctions file isn't quite to that point yet.  But, I think we're quickly going to be adding a lot more functions to it, so it makes sense.  Also, these functions have almost no interaction with each other across the file boundaries.  Just the exception helper that's in the Core file.

That said, it's also not a hill I care about enough to die on.  So I'd be glad to change it back to one file if you disagree.

Brant

jeffrymorris:
Brant -

Good work, thanks!

-Jeff

PR #100, opened 2016-01-04
Make internal documentation links relative

brantburnett:
Motivation
----------
Docs currently always link to master branch of the main couchbaselabs
repo.  This means that any version specific documentation in a specific
repo fork or in an older branch are lost when following the links.

Modifications
-------------
Changed all internal documentation links to be relative to the current
document path.

Results
-------
Following internal documentation links keeps the user in the same repo and
branch/tag.


jeffrymorris:
Thanks Brant!

PR #101, opened 2016-01-04
Support for IN operator on nested arrays via Contains

brantburnett:
Motivation
----------
N1QL supports testing nested arrays for values using the IN operator.
Contains is an equivalent LINQ function which can be supported.

Modifications
-------------
Implement ContainsResultOperator for nested array subqueries, mapping to
the IN statement in N1QL.

Throw an exception if an unsupported result operator is used for a query.
The current handling was to fail transparently without an exception,
ignoring the result operator.

Results
-------
The LINQ Contains method may be used against nested arrays within
Couchbase documents.


jeffrymorris:
Thanks Brant!

PR #103, opened 2016-01-05
Refactor integration tests for issue#94

jeffrymorris:
Motivation
----------
While somewhat expected, the integration tests run extremely slow. This
commit addresses a number of issues effecting the slowness and speeds up
the tests significantly.

Modifications
-------------
Made all queries limit the result set to one. Added additional assertions
to ensure only one item is returned. Refactored the ClusterHelper
initialization to a global TestSetup.cs class that will run first before
any other tests. Made all QueryTests.cs tests reuse the bucket and cluster
via ClusterHelper. QueryTests no longer derives from N1QLTestBase.

Results
-------
On my machine using a vagrants VM cluster, the integration tests went from
over 5 minutes to run to around 1.5 mins.


brantburnett:
@jeffrymorris,

Looks good to me.

Brant

Issue #107, opened 2016-01-15
Filtering by multiple values


RafaelMontegordoFelix:
Hi,
I'm trying to do the following:

I have multiple objects with this structure on couchbase.
{
  "someproperty": "sometext",
  "anotherproperty": "3",
  "filterproperty": 1,
}
{
  "someproperty": "sometext",
  "anotherproperty": "3",
  "filterproperty": 2,
}

I'm trying to do this query.
```
var MultipleFilterProperties = new List<long>(){2};
var query = context.Where(q => MultipleFilterProperties.Contains(q.filterproperty));
```

Regardless of the values that MultipleFilterProperties list have, the query return all items in couchbase, filter do not occur.

Is this a limitation?
It's possible to do this with linq2couchbase?

Thanks

brantburnett:
@RafaelMontegordoFelix 

I don't believe that this is currently supported.  We recently added [support for the IN keyword](https://github.com/couchbaselabs/Linq2Couchbase/commit/050ffb051f6015e77b5503f5c4434f8f52a89e43), but this was for checking for a single value in an array nested inside your document.  I think we clearly need to implement this situation, which is the inverse.

I will see what I can do to get this complete before our next release.

Thanks,
Brant

RafaelMontegordoFelix:
Thank you.

brantburnett:
@RafaelMontegordoFelix 

I was able to get this working without too much trouble.  It will be included in the next release, which is coming soon.  Or you may download the master branch and build it yourself if you need it sooner.

Thanks,
Brant

RafaelMontegordoFelix:
Nice, thank you very much.

PR #108, opened 2016-01-16
Issue #107: Support for subquerys against constant lists

brantburnett:
Motivation
----------
A desirable use case is to use the .Contains function against a static
list to test if a document attribute matches at least one item in the
list.  This can be implemented with the N1QL IN keyword.

Modifications
-------------
Relinq treats this is a subquery, with a constant value implementing
IEnumerable as the expression on the from clause.  Detect this case, and
render a static array to N1QL when it is encountered.  We can treat this
like any other array-type subquery, which includes support for the
.Contains method.

Results
-------
array.Contains(x) can now be used within a query, where array is constant
in relation to the query.  Note that array can be dynamically built before
the query, it is simply constant within the query's execution plan and
does not vary based on data returned within the query.

Also, due to the implementation approach, it also supports more advanced
subqueries against constant arrays.  For example, it is possible to filter
or perform select projections against the constant array before the
.Contains statement.  In the future, it might be worthwhile to optimize
these cases by performing filtering and projections before writing the
array to the N1QL query.  However, it is functional now.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
@@ -708,6 +708,27 @@ namespace Couchbase.Linq.IntegrationTests
         }
 
         [Test]
+        public void SubqueryTests_StaticArraySubqueryContains()
+        {
+            var bucket = ClusterHelper.GetBucket("beer-sample");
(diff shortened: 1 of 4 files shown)


jeffrymorris:
@brantburnett -

Nice work, thanks for the submission!

-Jeff

PR #109, opened 2016-01-16
LINQ-17: Support for UNION statements

brantburnett:
Motivation
----------
Currently, UNION statements between must be executed in memory, after
using the .AsEnumerator or .ToList methods.  This is much less efficient
than operating on the server in N1QL, especially since the two queries
must be run sequentially instead of concurrently.

Modifications
-------------
Added support for UnionResultOperator (UNION) and ConcatResultOperator
(UNION ALL).  Subqueries are parsed and added to the QueryPartsAggregator
in a new UnionParts collection.

Also implemented support for sorting after the UNION statement is
complete.  This required adjustments to the handling of sub queries in the
main from clause.  To reduce confusion, renamed the private GroupingStatus
enumeration to VisitStatus.

To support sorting on the union statement, adjustments were also made to
N1QlExtentNameProvider and N1QlExpressionTreeVisitor.VisitMember.  They
required support for the idea of accessing members of objects without
prefixing them with `Extent1`.

Results
-------
UNION and UNION ALL statements are now possible from LINQ.  Also supports
applying sorts after the UNION statement is complete.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
@@ -771,6 +771,52 @@ namespace Couchbase.Linq.IntegrationTests
         }
 
         [Test]
+        public void SubqueryTests_Union()
+        {
+            var bucket = ClusterHelper.GetBucket("beer-sample");
+            var context = new BucketContext(bucket);
+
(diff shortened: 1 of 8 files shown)


jeffrymorris:
Thanks Brant!

PR #110, opened 2016-01-17
Internalize Castle.Core and Remotion.Linq assemblies

brantburnett:
Motivation
----------
Reduce the number of dependencies required for the Couchbase.Linq
Nuget package.  Also, improve side-by-side compatibility with other
libraries which may use different versions of these assemblies that
aren't compatible.

Modifications
-------------
Added a Nuget dependency on ILRepack, which downloads the executable
before building.  Created a new solution configuration named
"RepackedRelease", which triggers the ILRepack process after the normal
build steps are complete.  The final assembly can be found in the
Src/Couchbase.Linq/bin/RepackedRelease folder.

The RepackedRelease build mode does not build the test assemblies.  This
is because they don't function cleanly, because they already access
internal classes in Couchbase.Linq and make use of Remotion.Linq and
Castle.Core classes.  Once repacked, the references then disagree with
each other.

Results
-------
The Couchbase.Linq assembly has both Castle.Core and Remotion.Linq
assemblies embedded within it.  When embedded, the assemblies are marked
as internal so they are not a part of the exposed API.

This means that internally the full functionality of both assemblies are
available for use.  But dependencies on these libraries don't need to be
added to the Nuget package definition, and they don't need to be installed
by the consuming project.  It also means that consumers can choose to use
other packages, such as Entity Framework, which may have dependencies on
different versions of these packages that are not compatible.

Notes
-----
Unfortunately, there wasn't a good way to run the unit and integration
tests against the repacked version of the assembly.  Therefore, for
testing purposes I used another project and consumed the repacked assembly
successfully.

### DIFF
--- a/Src/Couchbase.Linq/Couchbase.Linq.csproj
+++ b/Src/Couchbase.Linq/Couchbase.Linq.csproj
@@ -13,6 +13,7 @@
     <FileAlignment>512</FileAlignment>
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
     <RestorePackages>true</RestorePackages>
+    <ILRepackPath>$(SolutionDir)\packages\ILRepack.2.0.10\tools\ILRepack.exe</ILRepackPath>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


jeffrymorris:
Thanks Brant!

Issue #102, opened 2016-01-05
Count() called on IQueryable<T> with projection causes N1QL syntax error


jeffrymorris:
Is this a bug? Take a query:

var beers = from beer in context.Query<Beer>()
                                where (beer.Type == "beer")
                                select new { beer.Name, Year = N1QlFunctions.DatePart(beer.Updated,     N1QlDatePart.Year) };

var results = beers.Take(1);
Assert.AreEqual(1, results.Count());

This will error with:

 {"msg":"syntax error - at as","code":3000,"name":null,"sev":0,"temp":false}

The problem seems to be that the generated N1QL query contains a Count function like this:

SELECT COUNT(`Extent1`.`name` as `name`, DATE_PART_STR(`Extent1`.`updated`, 'year') as `year`) 
FROM `beer-sample` as `Extent1` 
WHERE (`Extent1`.`type` = 'beer') LIMIT 1

The query engine doesn't like the 'as' within the COUNT function:

cbq> SELECT COUNT(`Extent1`.`name` as `name`) FROM `beer-sample` as `Extent1` WHERE (`Extent1`.`type` = 'beer') LIMIT 1;
{
    "requestID": "bbab5d10-ee72-4d36-920c-74025734c6ef",
    "errors": [
        {
            "code": 3000,
            "msg": "syntax error - at as"
        }
    ],
    "status": "fatal",
    "metrics": {
        "elapsedTime": "1.194628ms",
        "executionTime": "1.133948ms",
        "resultCount": 0,
        "resultSize": 0,
        "errorCount": 1
    }
}

cbq> SELECT COUNT(`Extent1`.`name`) FROM `beer-sample` as `Extent1` WHERE (`Extent1`.`type` = 'beer') LIMIT 1;
{
    "requestID": "d97b04b6-6c42-448f-8bd5-09f150fdf963",
    "signature": {
        "$1": "number"
    },
    "results": [
        {
            "$1": 5891
        }
    ],
    "status": "success",
    "metrics": {
        "elapsedTime": "2.59933744s",
        "executionTime": "2.599289114s",
        "resultCount": 1,
        "resultSize": 34
    }
}

Could also be a query engine bug...or not, why include the project within the COUNT?

-Jeff

brantburnett:
@jeffrymorris 

[...]

PR #112, opened 2016-01-21
Issue #102: Aggregates after select projections

brantburnett:
Motivation
----------
If you apply a select clause in LINQ, and then follow this with an
aggregate such as .Count(), the generated N1QL is invalid.  It is
including syntax like "`a` as `a`, `b` as `b`" inside the COUNT()
function.

Modifications
-------------
Recognize when a new object select statement is used within an aggregate.
When encountered, use the JSON {...} new object syntax in N1QL.

Results
-------
N1QL created is now valid and runs successfully.  Note that using the JSON
new object syntax like COUNT({a: b}) is more valid than simply always
using COUNT(*).  This is because it will work with other operations like
COUNT(DISTINCT {a: b}).


jeffrymorris:
Thanks Brant

PR #114, opened 2016-01-23
Load document ids into proxies automatically

brantburnett:
Motivation
----------
For change tracking, it is important to know the document ID of a document
read from the database.  This ensures that changes saved back to the
database are saved back to the same ID.  Any approach that doesn't read
the document IDs at load time is relying on a key generation system that
is always 100% consistent.

Modifications
-------------
If proxy generation is enabled, and if the query being run will generate
proxies, then append an extra element to the select projection. This
element should return the document ID in the __id property.

Add __id as a property to the proxy objects via ITrackedDocumentNode, and
store it in the DocumentNode object.  This value can then be used by the
change tracking engine.

Note that "__id" as the property name doesn't really follow normal naming
conventions.  However, using the JsonProperty attribute would only work
with Newtonsoft.Json based deserialization.  Naming the property to match
the name in the N1QL query increases the likelihood of compatibility with
other deserialization engines.  Also, naming the property "__id" decreases
the likelihood of a naming conflict with the document attributes.

Results
-------
Document IDs are automatically read for any document being proxied for
change tracking, and made available to the change tracking system.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using Couchbase.Linq.Extensions;
 using Couchbase.Linq.IntegrationTests.Documents;
 using Couchbase.Linq.Proxies;
 using NUnit.Framework;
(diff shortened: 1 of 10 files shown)


jeffrymorris:
Thanks Brant!

PR #116, opened 2016-02-17
Handle GUIDs correctly in expression tree.

akorczynski:
This fixes the issue with using a Guid comparison in a where clause.  I went ahead and added a test and an additional test document class (Car) that has a Guid.  Didn't want to mess with any existing documents since I'm not familiar with the library.


brantburnett:
@jeffrymorris 

I've reviewed and this work looks good to me.

Thanks,
Brant

jeffrymorris:
@akorczynski @brantburnett Thanks!

PR #117, opened 2016-02-22
LINQ-27: Asynchronous LINQ query support

brantburnett:
Motivation
----------
Provide a transparent, easy to use mechanism for consumers to execute LINQ
queries against Couchbase asynchronously.  This should include support for
queries which use scalar result operations, such as .First(), .Sum(),
.Any(), etc.

Modifications
-------------
Created LinqQueryRequest as an extension of the SDK QueryRequest.  It
provides additional properties specific to the execution of queries
created for LINQ.  Refactored BucketQueryExecutor to use this construct
internally for query execution.

Created IBucketQueryExecutor to extend IQueryExecutor, which provides
extensions to execute the query asynchronously by providing a
LinqQueryRequest.  Extended BucketQueryable with the new interface
IBucketQueryExecutorProvider to give an method for internal consumers to
access the IBucketQueryExecutor.

Created ToQueryRequestExpressionNode, ToQueryRequestResultOperator,
LinqQueryRequestDataInfo, and LinqQueryRequestValue to allow for fake
query execution.  This fake query execution doesn't actually run against
the bucket, but instead creates the LinqQueryRequest and returns it.
This process can be easily triggered via
LinqQueryRequest.CreateQueryRequest.

Extended IQueryable with two overloads of ExecuteAsync, both of which
execute a query against a bucket asynchronously.  The first returns an
IEnumerable.  The second takes an additional lambda, and returns a scalar
result.  This is used to apply immediately executed result operators, such
as .First(), .Sum(), .Any(), etc.

In order to provide support for .Any() and .All(), refactored the handling
of result extraction for these operations.  Previously this was handled
directly in BucketQueryExecutor.ExecuteScalar, but this method isn't used
when executing asynchronously.  Now settings related to this are stored in
ScalarResultBehavior on LinqQueryRequest, and the calls to
ExecuteCollection and ExecuteCollectionAsync apply the configured behavior
automatically.

Results
-------
Any LINQ query against a bucket can now be executed asynchronously.  This
is done with only a very small change to the exposed API, the two
overloads of ExecuteAsync which extend IQueryable.

Additionally, the use of LinqQueryRequest internally lays the groundwork
for further extensions which support query parameters such as scan
consistency.


jeffrymorris:
Thanks @brantburnett !!! Nice work :)

PR #120, opened 2016-03-04
Support index consistency settings on LINQ queries

brantburnett:
Motivation
----------
As an SDK consumer, I'd like to be able to specify my index consistency
requirements when using LINQ.  This extends the functionality of the LINQ
client to include functionality that already exists for handwritten
queries.

Modifications
-------------
Added IQueryable<T> extensions for setting ScanConsistency and ScanWait.
These values are passed through to BucketQueryExecutor, which then applies
them to the resulting LinqQueryRequest.

Results
-------
Users can apply index consistency settings inline as they are building the
LINQ query.

Note: ScanVector is not implemented, because dynamic types are not
supported on extension methods.  I'm also not sure that Couchbase Server
supports this parameter yet.


jeffrymorris:
Thanks @brantburnett !!!

PR #121, opened 2016-03-06
Improved query exception handling

brantburnett:
Motivation
----------
Current exception handling during query execution can be unclear, and is
difficult to trap with filtered catch blocks.

Modifications
-------------
All errors during query execution are thrown as a CouchbaseQueryException.

If the Errors collection is present and populated, include it in the
Errors property of the CouchbaseQueryException.  If there is only one
error, its message becomes the exception message.  If there is more than
one error, a generic error message is used that refers the developer to
the Errors property.

If an Exception is returned instead, a generic error message is used and
the Exception becomes the InnerException of the CouchbaseQueryException.

If neither Errors or Exception is populated (which should not happen
unless there is a bug further up the line), the message indicates that
there was an unknown error.

Results
-------
Exceptions can be easily trapped by catching CouchbaseQueryException
specifically.  If trapped, the entire Errors collection is available in a
property, as is the Exception in the InnerException property.  The
exception message is more readable and informative.

### DIFF
--- a/Src/Couchbase.Linq/Couchbase.Linq.csproj
+++ b/Src/Couchbase.Linq/Couchbase.Linq.csproj
@@ -92,6 +92,7 @@
     <Compile Include="Clauses\UseKeysExpressionNode.cs" />
     <Compile Include="Clauses\UseKeysClause.cs" />
     <Compile Include="BucketContext.cs" />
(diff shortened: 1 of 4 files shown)


jeffrymorris:
Thanks Brant!

PR #125, opened 2016-03-21
Support UPDATE operator

oguimbal:
## Motivation

There is currently no support for the N1QL "UPDATE"  operator.
## Modifications

Adds 3 extension methods to IQueryable<T>:  Set, Unset and Execute.
## Results

User can now use update operator. Usage examples:

```
var newCustomers = query<User>()
      .Where(x=> x.Age > 20)
      .Set( x => x.CanHaveBeers == true )
      .Unset(x => x.WeirdVoice )
      .Select(x=>x.Name)
     .ToArray();
```

which will translate to N1QL:

```
 UPDATE default x SET x.canHaveBeers=true UNSET x.weirdVoice WHERE x.age > 20 RETURNING x.name
```

[see update reference](http://developer.couchbase.com/documentation/server/4.0/n1ql/n1ql-language-reference/update.html)

**nb1:** You may have noticed that i'm using "==" operator in expressions. That's just a trick because unfortunately, C# does not support assignations in expressions. (unless built manually with Expression.Assign)

**nb2:** You could also use .Execute() to not return user names, which saves bandwidth.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
@@ -1164,5 +1164,35 @@ namespace Couchbase.Linq.IntegrationTests
         }
 
         #endregion
+
+
+        [Test]
+        public void Update_Set_Unset_Execute()
+        {
+            var bucket = ClusterHelper.GetBucket("beer-sample");
+            var context = new BucketContext(bucket);
+
+            var what = context.Query<Beer>().Where(beer => beer.Type == "beer" && beer.Name == "21A IPA");
+
+            Assert.AreEqual("American-Style India Pale Ale", what.First().Style);
+
+            var beers = what.Unset(x => x.Style)
+                            .Select(x => new { x.Name, x.Style })
+                            .ToArray();
+
+            Assert.AreEqual(1, beers.Length);
+            Assert.AreEqual(null, beers[0].Style);
+            Assert.AreEqual("21A IPA", beers[0].Name);
+
+            // re-execute request
+            Assert.AreEqual(null, what.First().Style);
+
+            what.Set(x => x.Style == "American-Style India Pale Ale").Execute();
+
+
+            // re-execute request
+            Assert.AreEqual("American-Style India Pale Ale", what.First().Style);
+        }
+
     }
 }
--- a/Src/Couchbase.Linq.UnitTests/Couchbase.Linq.UnitTests.csproj
+++ b/Src/Couchbase.Linq.UnitTests/Couchbase.Linq.UnitTests.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
@@ -146,6 +146,7 @@
     <Compile Include="QueryGeneration\OrderByClauseTests.cs" />
     <Compile Include="QueryGeneration\SelectTests.cs" />
     <Compile Include="QueryGeneration\TakeAndSkipTests.cs" />
+    <Compile Include="QueryGeneration\UpdateClauseTests.cs" />
     <Compile Include="QueryGeneration\WhereClauseTests.cs" />
     <Compile Include="TestConfigurations.cs" />
   </ItemGroup>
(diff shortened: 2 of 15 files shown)


brantburnett:
So, I haven't done a code review, I've just looked at the API itself so far.  It's a very interesting concept, and I see it as a way to differentiate Couchbase with other platforms.  While everything update related could be done via a SELECT query followed by SDK update operations, this approach has performance benefits because it offloads more of the work to the query node.  It should also automatically make use of new Couchbase features like the subdocument API coming in 4.5, once it's added to the query node UPDATE statement logic.

Here are some thoughts I had about the API itself.

1. I think that using the equality operator is pretty confusing.  Doesn't really fit with the normal C# API approach.  Also, it will make it difficult for consumers to perform unit testing.  Instead, why not just make it consistent with the way that Unset was implemented.  Simply add a Set operation that acts like Unset, but takes an additional parameter containing the new value or an Expression to derive it.
2. N1QL also supports some more advanced logic for setters, such as logic to find the right element in an array to update.  It would be really cool to support this function.  But this could be added in the future, it should just involve more advanced parsing of the expression passed to Set or Unset.
3. @jeffrymorris is in the midst of creating a change tracking system, which will allow you to track document changes and submit them all when you're done with your unit of work.  I'm wondering how we could make these two update systems work well with each other.  Maybe divert the processing if change tracking is enabled, and just execute it as a SELECT query?  Then execute the requested changes in memory so they hit the change tracking engine.  Just thinking out loud here.

Thanks for the great submission!

Brant

jeffrymorris:
@oguimbal -

Very cool, thanks for submitting. Paralleling @brantburnett's comments, we need to think where this fits in with change tracking and how we can leverage some of the newer Couchbase API's so that's it's using the best tool for the job.

-Jeff

oguimbal:
Thanks :)

**1)** Regarding equality operator: That's what i've originally done. Ive seen two drawbacks:

  - As you have mentioned, two overloads must be created. One with a value, one with an expression. The syntax is OK using the overload with a value ( `.Set(x=>x.Member, value)` )  but a bit more painful when it comes to the expression one (i find that .Set(x => x.Member, x=>x.OtherMember + 5)` is a bit noisy). But anyway, that's quite a small one.

- A bit more annonying:  With those two overloads, intellisense is not smart enough to understand what we're doing... which just means that we're losing code completion on .Set expressions... I felt a bit inconfortable about this one, that's why I used the equality operator trick.

   I can switch back to this solution, though. But at least you will know its implications !

**2)** Yes, i've been a bit lazy about this one. But when it comes to choosing the API syntax, I thought it would be pretty natural to be:

```Query<T>()
         .SelectMany(x=>x.Children) // <=  This one is perfect for the job
         .Where(x=>x.HasCleanedHisRoom)
         .Set(x=>x.WillGetPresentForXmas == true) // or .Set(x=>x.WillGetPresentForXmas, true)  !
         .Execute()
```

i'll probably need it sometime, so i may implement it !

**3)** No sure what your tracking system looks like, but are you thinking about compiling the query predicates, and executing them against all tracked object in order to reflect the select query ?
My two cents on this question: I've been dealing a lot with caching and linq providers, and it's really complicated. Mainly because of two things:

- Query behavior replication: The compiled predicate must behave exactly like the one executed against the database. Just picture a where clause `x=>x.child.myProperty == null`. You obviously cannot compile and execute it like this without thinking having a NullReferenceException somewhere. And i'm not mentioning operators which may not behave the same (yes, i'm taliking about you, sql LIKE operator). 
You'll always have funny things you've never though about. Just picture this scenario: One of your employees had a children. So you've decided to add the property `int ChildCount{get;set;}` to your `Employee`class. Fortunately, noboy else have children, so even if no stored document has this "childCount" property, it's ok: Its filled with 0 by default.
Now lets update:
`` query<x>().Where(x=>x.ChildCount==0).Set(x=>x.Status = "May be single") ``  This one will **not** update anyting against the DB since no "age" property is set. But it will update everyone in your tracking engine system, making your DB and your memory instances inconsistent.

That's not the kind of hard-to-debug-and-possibly-debilitating issue you want to introduce in your provider without giving it some serious thougths.

- Performance: You can track hundreds of thousands of objects, all it takes is more memory...but executing a predicate on each one of them comes with a cost. There are some better performing solutions, though. But it will require a lot of dev, probably adding a proportially huge amount of bug.

oguimbal:
And yes, it's obviously better performing than loading=modifying=saving every document.

But I think what it's really better about is concurrency: load=modify=save gives room for two concurrent updates.

ex: Given the timeline:
Thread 1: ===(load x)===(modify: x.prop=123)===========(save x)===========
Thread 2: ========(load x)====(modify: x.otherprop=123)=========(save x)===

=> The first update will be overridden.

Using UPDATE may or may not prevent this kind of concurrency issue (i'm not sure if UPDATE statement provides any atomicity guarantee, i guess not if it's executed by index services)... but it surely removes network latency and a lot of network payload on big documents, therefore reducing risks of occurence of these issues.

jeffrymorris:
Hello @oguimbal -

I was wondering if you were going to do another revision without the equality operator trick? I can see this (SET/UNSET) being a useful feature.

-Jeff

oguimbal:
@jeffrymorris Yes I could, but I am a bit busy righ now. It could do it around september.
It's a pretty easy and straightforward change, though... so if someone wants it sooner, just have a look at the diff of this PR :)

jeffrymorris:
Created a ticket for tracking: https://issues.couchbase.com/browse/LINQ-29

jeffrymorris:
Closing due to inactivity. Feel free to ping me if necassary.

PR #127, opened 2016-04-04
LINQ-19: Add change-tracking to BucketContext

jeffrymorris:
Motivation
----------
This commit introduces experimental support for change tracking of
entities/POCOs loaded from documents stored in Couchbase server. By enabling
change tracking through a call to the BeginChangeTracking method on
BucketContext, each POCO or entity that corresponds to a document returned
by a query will be intercepted and proxied. The proxy contains logic for
detecting any modifications to a document and adding the corresponding
object to a mutated list. When SubmitChanges is called, the changed
documents will be added, updated or deleted on Couchbase Server. Change
tracking be disabled by calling the EndChangeTracking method; which will
remove any modified documents from the mutated list and subsequent queries
will not be intercepted and proxied.

Modifications
-------------
The BucketContext is now a subcriber for all proxy modifications when
change tracking is enabled; any changes to an object which is proxied will
be bubbled up to the BucketContext and stored in a mutated list. If a new
document is added that was not proxied it will be wrapped in a
NewDocumentWrapper class and treated as if it was proxied. Various other
methods of BucketContext were also modified to support the cases where
chnage tracking is enabled and when it is not enabled. In general, any
delete, add or mutation persistence will be deferred until SubmitChanges
is called.

Results
-------
When BegingChangeTracking is enabled all documents returned by a query
will be proxied. If a document is modified, the BucketContext will be
notified. When SubmitChanges it called, the documents that have been
changed or added will be persisted to Couchbase.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
@@ -19,10 +19,8 @@ namespace Couchbase.Linq.IntegrationTests
                 where x.Type == "beer"
                 select x;
 
-            foreach (var beer in query.Take(1))
-            {
-                Console.WriteLine(beer.Name);
-            }
+            var beer = query.FirstOrDefault();
+            Assert.IsNotNull(beer);
         }
 
         [Test]
@@ -32,10 +30,8 @@ namespace Couchbase.Linq.IntegrationTests
             var beers = from b in db.Beers
                 select b;
 
-            foreach (var beer in beers.Take(1))
-            {
-                Console.WriteLine(beer.Name);
-            }
+            var beer = beers.Take(1);
+            Assert.IsNotNull(beer);
         }
 
         [Test]
@@ -58,10 +54,9 @@ namespace Couchbase.Linq.IntegrationTests
         [Test]
         public void Query_EnableProxyGeneration_ReturnsProxyBeer()
         {
-            var db = new BucketContext(ClusterHelper.GetBucket("beer-sample"))
-            {
-                EnableChangeTracking = true
-            };
+            var db = new BucketContext(ClusterHelper.GetBucket("beer-sample"));
+
+            db.BeginChangeTracking();
 
             var query = from x in db.Query<Beer>()
                         where x.Type == "beer"
@@ -79,10 +74,8 @@ namespace Couchbase.Linq.IntegrationTests
         [Test]
         public void Query_EnableProxyGeneration_ReturnsProxyBeerWithId()
         {
-            var db = new BucketContext(ClusterHelper.GetBucket("beer-sample"))
-            {
-                EnableChangeTracking = true
-            };
+            var db = new BucketContext(ClusterHelper.GetBucket("beer-sample"));
+            db.BeginChangeTracking();
 
             const string documentId = "21st_amendment_brewery_cafe-21a_ipa";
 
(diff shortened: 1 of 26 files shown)


Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
```diff
@@ -19,10 +19,8 @@ public void Test_Basic_Query()
                 where x.Type == "beer"
                 select x;
 
-            foreach (var beer in query.Take(1))
-            {
-                Console.WriteLine(beer.Name);
-            }
+            var beer = query.Take(1);
+            Assert.IsNotNull(beer);
```
brantburnett: Jeff,

Looks like beer here is an IQueryable<Beer>, not an individual Beer, so IsNotNull will always be true.  I think it should use .FirstOrDefault() instead.

Src/Couchbase.Linq.UnitTests/BucketContextTests.cs
```diff
@@ -12,95 +15,334 @@ public class BucketContextTests
...
+
+        [Test]
+        public void Remove_When_DocId_Is_Not_Defined_Throw_DocumentIdMissingException()
+        {
+            //arrange
+            var brewery = new Brewery();
+            var bucket = new Mock<IBucket>();
+            var result = new Mock<IOperationResult<Brewery>>();
+            result.Setup(x => x.Status).Returns(ResponseStatus.Success);
+            bucket.Setup(x => x.Upsert(It.IsAny<string>(), It.IsAny<Brewery>())).Returns(result.Object);
```
brantburnett: Jeff,

I think this is supposed to be mocking Remove instead of Upsert here.

Brant

Src/Couchbase.Linq.UnitTests/BucketContextTests.cs
```diff
@@ -12,95 +15,334 @@ public class BucketContextTests
...
+
+        [Test]
+        public void EndChangeTracking_ChangeTrackingEnabled_Is_False()
+        {
+            //arrange
+            var bucket = new Mock<IBucket>();
+            bucket.Setup(x => x.Configuration).Returns(new ClientConfiguration().BucketConfigs.First().Value);
+
+            var ctx = new BucketContext(bucket.Object);
+            ctx.BeginChangeTracking();
```
brantburnett: Jeff,

Do you think we need another unit test for calling BeginChangeTracking twice, followed by a single EndChangeTracking, to ensure that change tracking is still operational?

Src/Couchbase.Linq.UnitTests/Proxies/IgnoreProxyFieldContractResolverTests.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Threading.Tasks;
+using Couchbase.Linq.Proxies;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+using NUnit.Framework;
+
+namespace Couchbase.Linq.UnitTests.Proxies
+{
+    [TestFixture]
+    public class IgnoreProxyFieldContractResolverTests
```
brantburnett: Jeff,

I think the IgnoreProxyFieldContractResolver and it's associated tests are leftovers, it doesn't appear to be referenced anywhere.

Src/Couchbase.Linq/Proxies/DocumentProxyGenerationHook.cs
```diff
@@ -51,7 +51,7 @@ public bool ShouldInterceptMethod(Type type, MethodInfo methodInfo)
             // Only proxy getters and setters for properties on the document
             // We must proxy getters or serializing the document back to Couchbase has problems
 
-            return methodInfo.IsSpecialName;
+            return methodInfo.IsSpecialName && methodInfo.Name.StartsWith("set_") || methodInfo.Name.StartsWith("get_");
```
brantburnett: Jeff,

I think there's an operator precedence issue here if you don't wrap the || clause in parentheses.

jeffrymorris:
@brantburnett  -

Thanks for the review, will push fixes tomorrow.

-Jeff

jeffrymorris:
@brantburnett -

Forced pushed the changes. 

-Jeff

jeffrymorris:
Merged and closed. This is an *experimental* feature - any issues should be addressed by future commits. Please use the[ project Jira ](https://issues.couchbase.com/projects/LINQ )or create an issue on this github project if find a bug or want support for a missing feature.

Issue #129, opened 2016-04-28
WhereClause culture problem


HeuHax:
I have run into a problem when doing this under a danish culture setting.
`var activities = DbContext.Context.Query<Location>().Where(x =>
     x.Geo.Lng > 56.5
 );`

Basically what is happening is that 56.5 is getting converted into 56,5 in the WhereClause and that makes the N1Ql query fail with the error "syntax error - at ,".

How do I fix or get around this ?

brantburnett:
Yeah, that definitely sounds like a bug.  We need to be rendering the numeric values using the invariant culture.

If your interested, you can submit a pull request for this problem, I think the solution would be around here:
https://github.com/couchbaselabs/Linq2Couchbase/blob/master/Src/Couchbase.Linq/QueryGeneration/N1QLExpressionTreeVisitor.cs#L453

Of course, I'd be happy to take a look at it myself, as well.

In the meantime, as a workaround you should be able to avoid the problem like this:

```csharp
var cultureCache = System.Threading.Thread.CurrentThread.CurrentCulture;
System.Threading.Thread.CurrentThread.CurrentCulture = System.Globalization.CultureInfo.InvariantCulture;
try
{
    // execute the query here
    // note that the query executes when you enumerate it or call .ToList(), not when you build it
}
finally
{
    System.Threading.Thread.CurrentThread.CurrentCulture = cultureCache;
}
```

HeuHax:
You are exactly right!
It's right at line 519:
`_expression.AppendFormat(CultureInfo.InvariantCulture, "{0}", namedParameter.Value);`

I guess this should be added everywhere where necessary, but I have only encountered the problem here so I don't know if aywhere else is necessary.

For now I say thanks !  :)

PR #130, opened 2016-04-30
Issue #129: Ensure N1QL numbers use period as decimal separator

brantburnett:
Motivation
----------
In cultures with a comma decimal separator, including a constant in your
query that has decimal precision will cause an invalid N1QL query.  N1QL
expects numbers to always use a period decimal separator.

Modifications
-------------
When serializing numeric constants to a N1QL query, always convert them to
a string using the Invariant culture.

Created a unit test to prevent future regressions.

Results
-------
N1QL queries with decimal constants will now work with any culture
setting, without requiring special workarounds by consumers.


jeffrymorris:
Thanks Brant!

PR #132, opened 2016-05-03
Exposed IBucket from IBucketContext

mgroves:
I thought it might be useful to be able to get to the underlying IBucket from a BucketContext: to drop down to lower-level functionality when necessary.

I also added an integration test that uses the exposed Bucket (it just runs a raw N1QL query).

Please let me know what you think.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
@@ -380,7 +380,27 @@ namespace Couchbase.Linq.IntegrationTests
             // execute N1QL to update the beer
             var newBeerName = Guid.NewGuid().ToString();
             var n1ql = string.Format("UPDATE `beer-sample` USE KEYS '{0}' SET name = '{1}';", beer.Id, newBeerName);
-            db.Bucket.Query<int>(n1ql);
+            db.Execute(n1ql);
+
+            // get the beer back out to make sure it was updated correctly
+            var beerAgain = db.Query<Beer>().First(b => N1QlFunctions.Meta(b).Id == beer.Id);
+
+            Assert.That(beerAgain.Name, Is.EqualTo(newBeerName));
+        }
+
+        [Test]
+        public void Underlying_Bucket_can_be_used_to_execute_N1QL_query_directly_with_parameterization()
+        {
+            // setup bucket context
+            var db = new BucketContext(ClusterHelper.GetBucket("beer-sample"));
+
+            // get a beer from the db
+            var beer = db.Query<Beer>().Select(b => N1QlFunctions.Meta(b)).First();
+
+            // execute N1QL to update the beer, use parameterization instead of string.Format
+            var newBeerName = Guid.NewGuid().ToString();
+            var n1ql = "UPDATE `beer-sample` USE KEYS $1 SET name = $2;";
+            db.Execute(n1ql, beer.Id, newBeerName);
 
             // get the beer back out to make sure it was updated correctly
             var beerAgain = db.Query<Beer>().First(b => N1QlFunctions.Meta(b).Id == beer.Id);
--- a/Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
@@ -379,7 +379,7 @@ namespace Couchbase.Linq.IntegrationTests
 
             // execute N1QL to update the beer
             var newBeerName = Guid.NewGuid().ToString();
-            var n1ql = $"UPDATE `beer-sample` USE KEYS '{beer.Id}' SET name = '{newBeerName}';";
+            var n1ql = string.Format("UPDATE `beer-sample` USE KEYS '{0}' SET name = '{1}';", beer.Id, newBeerName);
             db.Bucket.Query<int>(n1ql);
 
             // get the beer back out to make sure it was updated correctly
--- a/Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
@@ -367,6 +367,26 @@ namespace Couchbase.Linq.IntegrationTests
             var doc = ClusterHelper.GetBucket("beer-sample").GetDocument<Beer>(((ITrackedDocumentNode)beer).Metadata.Id);
             Assert.AreNotEqual(beer.Abv, doc.Content.Abv);
         }
+
+        [Test]
+        public void Underlying_Bucket_can_be_used_to_execute_N1QL_query_directly()
+        {
+            // setup bucket context
+            var db = new BucketContext(ClusterHelper.GetBucket("beer-sample"));
+
+            // get a beer from the db
+            var beer = db.Query<Beer>().Select(b => N1QlFunctions.Meta(b)).First();
+
+            // execute N1QL to update the beer
+            var newBeerName = Guid.NewGuid().ToString();
+            var n1ql = $"UPDATE `beer-sample` USE KEYS '{beer.Id}' SET name = '{newBeerName}';";
+            db.Bucket.Query<int>(n1ql);
+
+            // get the beer back out to make sure it was updated correctly
+            var beerAgain = db.Query<Beer>().First(b => N1QlFunctions.Meta(b).Id == beer.Id);
+
+            Assert.That(beerAgain.Name, Is.EqualTo(newBeerName));
+        }
     }
 }
 
--- a/Src/Couchbase.Linq.IntegrationTests/Couchbase.Linq.IntegrationTests.csproj
+++ b/Src/Couchbase.Linq.IntegrationTests/Couchbase.Linq.IntegrationTests.csproj
@@ -24,6 +24,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
+    <LangVersion>5</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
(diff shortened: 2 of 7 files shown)


Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
```diff
@@ -367,6 +367,26 @@ public void SubmitChanges_WhenDocsAreModifiedAndEndChangeTracking_DocumentIsNotM
...
+        {
+            // setup bucket context
+            var db = new BucketContext(ClusterHelper.GetBucket("beer-sample"));
+
+            // get a beer from the db
+            var beer = db.Query<Beer>().Select(b => N1QlFunctions.Meta(b)).First();
+
+            // execute N1QL to update the beer
+            var newBeerName = Guid.NewGuid().ToString();
+            var n1ql = $"UPDATE `beer-sample` USE KEYS '{beer.Id}' SET name = '{newBeerName}';";
```
brantburnett: @mgroves 

We're not yet supporting C# 6.0 language features in this project yet, it should still be able to build in VS 2013.

Brant

mgroves:
Doh, sorry about that. I'll change it to use string.Format

Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
```diff
@@ -367,6 +367,26 @@ public void SubmitChanges_WhenDocsAreModifiedAndEndChangeTracking_DocumentIsNotM
...
+            // setup bucket context
+            var db = new BucketContext(ClusterHelper.GetBucket("beer-sample"));
+
+            // get a beer from the db
+            var beer = db.Query<Beer>().Select(b => N1QlFunctions.Meta(b)).First();
+
+            // execute N1QL to update the beer
+            var newBeerName = Guid.NewGuid().ToString();
+            var n1ql = string.Format("UPDATE `beer-sample` USE KEYS '{0}' SET name = '{1}';", beer.Id, newBeerName);
+            db.Bucket.Query<int>(n1ql);
```
jeffrymorris: In this case Save(...) which uses K/V memcached protocol is faster/more efficient than going through N1QL over HTTP + all of the steps in-between before finally updating the document (using the same memcached protocol n the server). N1QL's only advantage here is SQL like syntax - but the path to get here is more complex. In fact, the motivation behind the Linq2CB was to abstract the N1QL syntax (which you cannot do 100% hence the N1QL helper functions) and make it easier for Linq2SQL and EF users to transition over to CB.

Src/Couchbase.Linq.IntegrationTests/Couchbase.Linq.IntegrationTests.csproj
```diff
@@ -24,6 +24,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
+    <LangVersion>5</LangVersion>
   </PropertyGroup>
```
jeffrymorris: Ah, LangVersion attribute...this is nice, but probably should be a separate commit.

Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
brantburnett: @jeffrymorris 

While the specific test being executed isn't really advantageous, I do like the support for this concept.  In particular, if the UPDATE statement was using a WHERE clause that updates multiple documents instead of USE KEYS.  In this case it will be more efficient to let the Query server handle reading and changing each document instead of on the client side.  On the client side, you would be forced to run a SELECT, then loop through the reuslts and run deletes.  This adds transferring all of the data from the Query node to the client unnecessarily.

Brant

Src/Couchbase.Linq/BucketContext.cs
```diff
@@ -71,7 +73,7 @@ public IQueryable<T> Query<T>()
         /// </value>
         public string BucketName
         {
-            get { return _bucket.Name; }
+            get { return Bucket.Name; }
         }
```
jeffrymorris: Instead of directly exposing the IBucket implementation here, why not instead add a method on IBucketContext  with a signature like this:

    public void/int Execute(string statement);

This will abstract out the underlying SDK semantics making it easier to "maintain" later on. 

Also, note that the two objects have very different lifespans: the IBucketContext impl is an ephemeral object and the CouchbaseBucket is a long-lived, application scoped object...this may cause confusion later on.

Src/Couchbase.Linq.IntegrationTests/BucketContextTests.cs
jeffrymorris: Yes, I didn't emote that but that is a very big advantage of UPDATE which you cannot do easily through K/V.

-Jeff

mgroves:
Whoops, I meant to make those csproj changes a separate commit.

I'll take a crack at making an Execute method

brantburnett:
So, throwing in my two cents here, I like the idea of just offering Execute to run update queries, instead of exposing the bucket.  However, what about things like RYOW using consistency levels, etc?  Maybe we should have two overloads?  One the shortcut method that this shows, and another that accepts an IQueryRequest to send more advanced options?

Also, I think squashing this down to one commit is probably preferable.  That will help keep the commit history cleaner when it's merged.

Brant

jeffrymorris:
@brantburnett - Hmm, i wonder how RYOW applies here...its surely an edge case, in that it doesn't make sense to follow an insert with an update does it? I am not opposed to it, however...

brantburnett:
@jeffrymorris 

You make a fair point about RYOW.  Maybe if you were running multiple update statements, and wanted the second not to run until after the first was indexed?  But definitely getting edgy.

There are other things, like timeout settings, on IQueryRequest which might apply, though.  Maybe worth thinking about in the future.

Brant

brantburnett:
@mgroves and @jeffrymorris

The more I'm thinking about this change, the more I think that maybe the original approach (adding a Bucket property to BucketContext) might be the best.  @mgroves, feel free to kick me now.  :)

Here are my reasons:

1.  We're already passing the IBucket to the constructor of Bucket, so the concept of a BucketContext having an IBucket is already part of the API.  So having a simple, read only shortcut to retrieve the IBucket originally provided makes sense.
2.  Using a Bucket property makes for a simpler API.  It provides a method to get to the IBucket and perform any advanced operation you want to, without complicating the IBucketContext interface with lots of methods over time.  Of course, if you just know the IBucket that was passed to the constructor you can use it directly.  But if you're using dependency injection, accessing the IBucket from the BucketContext is much simpler.
3.  For the specific query example which prompted this change, the Bucket property provides access to the full suite of functionality in the QueryRequest object.  As new features continue to be added, it will continue to receive these features transparently without further development on the BucketContext.  The same applies to any other features of the IBucket that might be required by a consumer.

Thoughts?

Brant

jeffrymorris:
@brantburnett @mgroves -

I have a couple of reservations here:

1. `IBucket` would now be coupled to `IBucketContext` - that is a hard interface dependency as opposed to a ctor. Imagine in the future we wanted to create a new `IBucketContext` implementation (bare with me: `EvenBetterThanTheOldBucketContext`) that didn't have dependencies on `IBucket`, we would have to live with throwing a `NotSupportedException` when the property is accessed. 
2. We now have widened the scope of the Linq2Couchbase API significantly - imagine people posting issues and bugs on this project because they have "dot operator-ed" themselves to the View API: `bucketContext.Query(new ViewQuery())` or `bucketContext.GetManager().CreateManager('boo", "foo").ListIndexes().First().Name;`. See [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) aka ["Don't talk to strangers"](https://en.wikipedia.org/wiki/Law_of_Demeter).
3. Linq2Couchbase was intended to be a _layer over the SDK_ - a very opinionated layer which drives the user down a specific path wrt usage and that hides complexity as much as possible. The abstraction was meant to be opaque to the user so we can use various API's defined by `IBucket` to "put together" common solutions for your everyday data needs.
4. I **do** think we should provide overloads that take `QueryRequest`.

-Jeff

brantburnett:
@jeffrymorris 

Okay, you make some good points there.

I guess one option would be do only add the `Bucket` property to the `BucketContext`, and not to the `IBucketContext` interface.  But that limits the usability in the dependency injection scenario.  If you're injecting the interface, you'd need to typecast to the class, etc.

But I kinda see your point here.  If you look at Entity Framework, they accept a `SqlConnection` as a constructor parameter when making a `DbContext`, but don't really expose it after that (at least, it's difficult to get to).  And they put functions similar to the `Execute` function we're discussing within a `Database` property on the `DbContext`.

Perhaps, if we're going down this road, we should consider putting the `Execute` function inside a child class like EF.  That might at least help with organization some, and keep from cluttering the main functionality in the `BucketContext` over time.

Maybe add a `IBucketContext.DataStore` property that returns an `IDataStore` interface that has the Query overloads?  Then make an internal `BucketDataStore` class that takes an `IBucket` as a constructor parameter, and is automatically placed in the `BucketContext.DataStore` property.  This preserves the option to implement `EventBetterThanTheOldBucketContext` that doesn't use `IBucket` in the future.

For the `Execute` function this could be overcomplicating it a bit.  But I think it might help in the future, and once we do something here we're pretty much stuck with it, so it's best to think about what we're doing down the road.

On a side note, we probably want to look at having `Execute` return values instead of void.  Maybe `IQueryResult` and/or `IQueryResult<T>`.  I can see wanting some of the data points like the number of documents affected for updates.  Also, I can see this being used to run SELECT queries that, for whatever reason, can't be generated by LINQ.  That's one of the use cases for the similar functions in Entity Framework.

Brant

jeffrymorris:
+1

brantburnett:
@mgroves 

How do you feel about this plan?  And is it something you'd like to implement?  Or would you prefer to pass off the work you've done so far and let someone else do all this extra stuff we're discussing?

Thanks,
Brant

mgroves:
It sounds like a good plan to me; I'd be glad to take a crack at it when I get some time.

mgroves:
Closing out this pull request in favor of a different one

PR #133, opened 2016-05-07
Support for Couchbase 4.5 index joins

brantburnett:
Motivation
----------
Couchbase 4.5 allows for flexible commutative joins, aka index joins.  For
more information, see
http://developer.couchbase.com/documentation/server/4.5-dp/flexible-join-n1ql.html.

Modifications
-------------
Added Versioning namespace, which defines a provider for determining and
caching the version number for a cluster.  This includes a suite of unit
and integration tests.

Updated the query generation system to output index joins when detected in
the LINQ query, but only if the cluster is on version 4.5 or later.  If
the cluster isn't on version 4.5, a NotSupportedException is thrown which
is consistent with previous behaviors.

Results
-------
Index joins can now be used against 4.5 clusters.

The behavior is largely unaffected for clients running against 4.0 and 4.1
clusters.  Differences include additional logs, a slight change to an
exception message, and a miniscule delay on the first query while the
version number of the cluster is determined.

### DIFF
--- a/Src/Couchbase.Linq/Couchbase.Linq.csproj
+++ b/Src/Couchbase.Linq/Couchbase.Linq.csproj
@@ -81,6 +81,7 @@
     <Reference Include="System.Configuration" />
     <Reference Include="System.Core" />
     <Reference Include="System.Net.Http" />
+    <Reference Include="System.Net.Http.WebRequest" />
     <Reference Include="System.Runtime.Serialization" />
     <Reference Include="System.Xml.Linq" />
     <Reference Include="System.Data.DataSetExtensions" />
@@ -192,6 +193,10 @@
     <Compile Include="QueryParserHelper.cs" />
     <Compile Include="UnixMillisecondsDateTime.cs" />
     <Compile Include="Utils\ExceptionMsgs.cs" />
+    <Compile Include="Versioning\DefaultVersionProvider.cs" />
+    <Compile Include="Versioning\FeatureVersions.cs" />
+    <Compile Include="Versioning\IVersionProvider.cs" />
+    <Compile Include="Versioning\VersionProvider.cs" />
   </ItemGroup>
   <ItemGroup>
     <None Include="app.config" />
--- a/Src/Couchbase.Linq/Versioning/DefaultVersionProvider.cs
+++ b/Src/Couchbase.Linq/Versioning/DefaultVersionProvider.cs
@@ -1,0 +1,200 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.Security;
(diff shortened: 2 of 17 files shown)


Src/Couchbase.Linq/Couchbase.Linq.csproj
```diff
@@ -81,6 +81,7 @@
     <Reference Include="System.Configuration" />
     <Reference Include="System.Core" />
     <Reference Include="System.Net.Http" />
+    <Reference Include="System.Net.Http.WebRequest" />
```
jeffrymorris: I thought we were doing away with WebRequest/WebResponse in favor of HttpClient? https://issues.couchbase.com/browse/NCBC-1098

brantburnett: We are, this is a bit of an oddity in HttpClient.  It is a set of extensions for HttpClient that adds some of the missing features that WebRequest had.  Specifically in this case, the ability to add a custom certificate validator.

If we add a .Net Core version of the LINQ client eventually, we may have to reevaluate because this assembly doesn't currently have a .Net Core build.  But I think Relinq is going to be our biggest blocker there.

jeffrymorris: Thanks for clarifying!

Src/Couchbase.Linq/Versioning/DefaultVersionProvider.cs
```diff
@@ -0,0 +1,200 @@
...
+            {
+                // Unwrap the aggregate exception
+                throw new HttpRequestException(ex.InnerException.Message, ex.InnerException);
+            }
+        }
+
+        private static bool ServerCertificateValidationCallback(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors)
+        {
+            Log.Info(m => m("Validating certificate: {0}", sslPolicyErrors));
+            return true;
```
jeffrymorris: This should actually be `return sslPolicyErrors == SslPolicyErrors.None;`

brantburnett: Yeah, you're right.  I just copied it from the main SDK at https://github.com/couchbase/couchbase-net-client/blob/master/Src/Couchbase/Configuration/Server/Providers/Streaming/HttpServerConfig.cs#L153.

Is that something we need to fix there as well, or was it intentional?

Brant

jeffrymorris: Yeah, needs to be fixed there as well. Originally it was intentional, but we should be fail if there are policy errors. 

Jeff

jeffrymorris:
Thanks Brant!

PR #135, opened 2016-05-14
Support for index-based NEST operations

brantburnett:
Motivation
----------
Couchbase Server 4.5 adds the option to perform NEST operations using keys
on the child document instead of a list of keys on the parent document.

Modifications
-------------
Represent index nest operations using the equivalent group join LINQ
operation.  Because group joins are also used to represent LEFT OUTER JOIN
operations, we must wait to see if an AdditionalFromClause was applied
that flattens the group join before we assume it is a NEST operation.

If there are predicates applied to the inner sequence before the NEST
operation, these are applied using an ARRAY filter and LET clause after
the NEST operation is complete.

Results
-------
LEFT OUTER NEST operations can now be performed using an index join.
There currently is no equivalent in LINQ for an INNER NEST using an index
join.  Since it can be emulated with a where clause, I'm not sure if
support is required at this time.

If the cluster is not running Couchbase Server 4.5, NotSupportedExceptions
are thrown, as before.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
@@ -587,7 +587,7 @@ namespace Couchbase.Linq.IntegrationTests
         public void JoinTests_InnerJoin_IndexJoin()
         {
             // To make required index:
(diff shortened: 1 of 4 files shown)


jeffrymorris:
Nice work, thanks Brant!

PR #136, opened 2016-05-16
Set LangVersion to C# 5 for all projects

brantburnett:
Motivation
----------
This solution is not yet supporting C# 6 language features.  We don't want
them to be suggested by VS or Resharper.

Modifications
-------------
Add LangVersion 5 tag to Debug and Release builds for all projects.  Fix
one example in Versioning of a C# 6 language feature.

Results
-------
Only C# 5 features are in use or suggested.


mgroves:
whoops, I think I just made the exact same pull request :P

jeffrymorris:
Thanks!

PR #137, opened 2016-05-16
created BucketDataStore and added to BucketContext

mgroves:
Refactoring into BucketDataStore according to our previous discussion

### DIFF
--- a/Src/Couchbase.Linq/BucketContext.cs
+++ b/Src/Couchbase.Linq/BucketContext.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Reflection;
@@ -11,7 +10,6 @@ using Couchbase.Linq.Filters;
 using Couchbase.Linq.Metadata;
 using Couchbase.Linq.Proxies;
 using Couchbase.Linq.Utils;
-using Remotion.Linq.Clauses;
 
 namespace Couchbase.Linq
 {
@@ -32,6 +30,8 @@ namespace Couchbase.Linq
         /// </summary>
         public bool ChangeTrackingEnabled { get { return _beginChangeTrackingCount > 0; } }
 
+        public IDataStore DataStore { get; private set; }
+
         /// <summary>
         /// Creates a new BucketContext for a given Couchbase bucket.
         /// </summary>
@@ -39,6 +39,7 @@ namespace Couchbase.Linq
         public BucketContext(IBucket bucket)
         {
             _bucket = bucket;
+            DataStore = new BucketDataStore(bucket);
         }
 
         /// <summary>
--- a/Src/Couchbase.Linq/BucketDataStore.cs
+++ b/Src/Couchbase.Linq/BucketDataStore.cs
@@ -1,0 +1,28 @@
+﻿using Couchbase.Core;
+using Couchbase.N1QL;
+
+namespace Couchbase.Linq
+{
+    /// <summary>
+    /// A set of lower-level methods for other bucket-related functionality
+    /// </summary>
+    internal class BucketDataStore : IDataStore
+    {
+        private readonly IBucket _bucket;
+
+        internal BucketDataStore(IBucket bucket)
+        {
+            _bucket = bucket;
+        }
+
+        /// <summary>
+        /// Execute a N1QL function. Optionally specify parameters
+        /// </summary>
+        public IQueryResult<T> Execute<T>(string statement, params object[] parameters)
+        {
+            var query = new QueryRequest(statement)
+                .AddPositionalParameter(parameters);
+            return _bucket.Query<T>(query);
+        }
+    }
+}
--- a/Src/Couchbase.Linq.IntegrationTests/BucketDataStoreTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/BucketDataStoreTests.cs
@@ -1,0 +1,55 @@
+﻿using System;
+using System.Linq;
+using Couchbase.Linq.IntegrationTests.Documents;
+using NUnit.Framework;
+
+namespace Couchbase.Linq.IntegrationTests
+{
+    /// <summary>
+    /// These tests don't necessary represent "good practices" in general use
+    /// (e.g. concatenation instead of parameterization, using N1QL update on a single document)
+    /// </summary>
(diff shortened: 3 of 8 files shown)


Src/Couchbase.Linq/BucketContext.cs
```diff
@@ -32,13 +30,16 @@ public class BucketContext : IBucketContext, IChangeTrackableContext
...
+        public IDataStore DataStore { get; private set; }
+
         /// <summary>
         /// Creates a new BucketContext for a given Couchbase bucket.
         /// </summary>
         /// <param name="bucket">Bucket referenced by the new BucketContext.</param>
         public BucketContext(IBucket bucket)
         {
             _bucket = bucket;
+            DataStore = new BucketDataStore(bucket);
```
jeffrymorris: Not to be a PITA, but upon seeing this I am not sure if we need a separate IDataStore abstraction; `Execute(...)` might be better off just being a member of IBucketContext. In general wrapping the IBucket in the ctor is going to be a bad decision later on. Perhaps a separate ctor might make it better: 

`public BucketContext(IDataStore dataStore)` 

and 

`public `BucketContext(IBucket bucket) : this(new BucketDataStore(bucket)`

But, every where else in `BucketContext` we use the `IBucket` implementation itself...it seems like duplication of responsibility and may make mocking/testing more complex (one more interface/layer to mock). Additionally, we're losing the ability to manipulate `QueryRequest` parameters as well; we should have an overload of `Execute` to handle  that as well or provide (or design a path towards) an equivalent  so that these values can be set. A good example of a value people will want to set is QueryRequest.Timeout.

-Jeff

Src/Couchbase.Linq/BucketDataStore.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        /// <summary>
+        /// Execute a N1QL function. Optionally specify parameters
+        /// </summary>
+        public IQueryResult<T> Execute<T>(string statement, params object[] parameters)
+        {
+            var query = new QueryRequest(statement)
+                .AddPositionalParameter(parameters);
+            return _bucket.Query<T>(query);
```
jeffrymorris: Unfortunately there is no way to modify any of the `QueryRequest` properties (like Timeout or ScanConsistency) if we go this route...at a minimum we should provide an overload which takes a `QueryRequest` or equivalent. Perhaps we need a public wrapper for QueryRequest configuration properties exposed somewhere?

Src/Couchbase.Linq/BucketContext.cs
brantburnett: @jeffrymorris 

I'm not sure that I totally agree with the idea of it becoming more complicated.  I think it actually simplifies things by providing a separation of concerns.  Core features are provided in BucketContext, and work by access the IBucket directly.  Advanced direct-access features are in IBucketDataStore, which in turn directly access the IBucket.

I think it also makes testing simpler.  If you're trying to mock IBucketContext, you can pretty easily choose to just not mock the IBucketDataStore if the code you're testing doesn't use the advanced direct-access features like Execute.  But if moved the Execute function into IBucketContext it's then harder to see that separation of concerns and not mock the functions.  

I do agree about the QueryRequest, though.

Brant

Src/Couchbase.Linq/BucketDataStore.cs
brantburnett: On another note, we might want to change the return type to directly return IEnumerable<T> and throw CouchbaseExceptions on errors.  This would be more consistent with the approach in the rest of the library.

jeffrymorris:
@mgroves are you still working on this?

-Jeff

mgroves:
I haven't looked at it in a while, no.

jeffrymorris:
Would like to check with @brantburnett on whether or not we would like include this in a future release.

-Jeff

brantburnett:
@jeffrymorris 

I think I still like the concept behind it.  When using dependency injection, having access to the bucket to run other queries can be very useful.  For that matter, having access to all bucket functionality would be useful.

I think we just need to figure out the approach that offers the best balance between ease of use, complication factor, maintainability, and future-proofing.  That's the tricky part.

Honestly, I think now I lean towards some method to simply get the IBucket object from the context.  This gives us access to the entire Couchbase feature set.  But I agree, I don't think we want to lock ourselves into the IBucket paradigm, which could change in the future.

However, the consumer will know what type of "backend" they are using.  So if we just provide access to the IBucket via a non-specific type, then the consumer could typecast it to get what they want.  For example, we could make IDataStore have no members, and you must type cast to BucketDataStore to use it, which then as a property that returns the IBucket.

Alternatively, we could still offer a simpler Execute method, ala the current pull request, on IDataStore.  Then add a BackingStore property that returns the IBucket, but cast as a plain object.  Then we have both a simple, agnostic way to run ad-hoc queries, combined with a way to do a typecast to get the IBucket if desired.

Brant

MikeGoldsmith:
To throw my oar in:

I don't see the problem in having a discrete get-only property that exposes the IBucket on BucketContext. BucketContext **is** a Couchbase implementation.

I think is works because:
1) It provides easy access to execute to use all the IBucket functionality, without having to re-wrap in the IDataStore
2) It's easily testable because it'll be a get-only interface property. All the IBucket functionality is tested in the main library
3) It's a very low-impact change and means you don't have to maintain & pass around references to both the IBucketContext and IBucket

@brantburnett @jeffrymorris @mgroves

brantburnett:
@MikeGoldsmith 

That seems fine to me.  The idea behind adding more complexity was to abstract away the IBucket somewhat for futureproofing.  But we already take IBucket in the BucketContext constructor, so I'm not sure how much value that abstraction would add.  If we ever have something other than an IBucket in the SDK, we'll probably have to make a new context type to support it anyway.

MikeGoldsmith:
I agree with the new context, or at least extensive changes if were to change the IBucket definition significantly.

@jeffrymorris How do you feel about this?

jeffrymorris:
@brantburnett @MikeGoldsmith -

Note that we have already changed the N1QL API significantly by adding Cluster Level Query (CLQ), but even with CLQ we just delegate to a random bucket in the credentials/config.

I am fine either way, I like the idea of abstracting the data store, but seems like an overkill since this is a data stores specific Linq provider. 

-Jeff

MikeGoldsmith:
Closing this PR in favour of [PR 177](https://github.com/couchbaselabs/Linq2Couchbase/pull/177).

PR #138, opened 2016-05-16
enforce C# 5 via csproj, updated a static initializer to use constructor

mgroves:
This will enforce the use of C# 5 via csproj files.

I also removed a use of a C# 6 language feature (see https://github.com/couchbaselabs/Linq2Couchbase/commit/0f17237e1f3ca07e680981f145ad6a3c2d5165a8)


mgroves:
I think this is a dupe of https://github.com/couchbaselabs/Linq2Couchbase/pull/136

jeffrymorris:
@mgroves I ended up using @brantburnett PR since it was first in the queue. Thanks for the submission, but we cannot use both so I am closing this one.

-Jeff

PR #144, opened 2016-06-28
Support for USE INDEX statements in queries

brantburnett:
Motivation
----------
Allow LINQ based query systems to supply index hints to the N1QL query
engine for improved performance.

Modifications
-------------
Added a new IQueryable extension, UseIndex, which applies the clause to
the query.  Accepts two possible parameters, the name of the index and the
optional type of the index.

Added constants for currently known index types under the N1QlIndexType class.

Added IsValidKeyword to N1QlHelpers to test for and prevent N1QL injection
attacks using invalid index types.

Added unit and integration tests for the new feature, plus documentation.

Results
-------
Users may now improve query performance by providing index hints when
using LINQ generated queries.  Both GSI and VIEW index types are
supported.  Because this support is provided via string constants, we are
already futureproofed for index types that might be added in future
versions of Couchbase Server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ NuGet will install the package and all dependencies. Once you have the resolved 
 - [Array Filtering, Projections, and Sorting](docs/array-filtering-projections.md)
 - [Grouping and Aggregation](docs/grouping-aggregation.md)
 - [The UseKeys Method](docs/use-keys.md)
+- [The UseIndex Method](docs/use-index.md)
 - [JOINing Documents](docs/joins.md)
 - [NESTing Documents](docs/nest.md)
 - [UNNESTing Documents](docs/unnest.md)
--- a/Src/Couchbase.Linq.IntegrationTests/App.config
+++ b/Src/Couchbase.Linq.IntegrationTests/App.config
@@ -9,6 +9,10 @@
       <section name="couchbase" type="Couchbase.Configuration.Client.Providers.CouchbaseClientSection, Couchbase.NetClient" />
     </sectionGroup>
   </configSections>
+  <appSettings>
+    <add key="adminusername" value="Administrator" />
+    <add key="adminpassword" value="password" />
+  </appSettings>
   <couchbaseClients>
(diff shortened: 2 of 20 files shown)


brantburnett:
@jeffrymorris 

I'd love some feedback on the way I handled GSI and VIEW index types in this one.  I was torn between using enumerations or strings.  Enumerations are easier to consume, but strings are more futureproof.

Brant

jeffrymorris:
Happy 4th Brant!

I did quick review and here are my notes:

-  The UseIndex unit tests fail initially until you create the proper index manually (" CREATE INDEX brewery_id ON `beer-sample` (brewery_id)"). Not a huge deal, but eventually we should script this so that the indexes are created when the tests run (at the expense of slower tests). The reason is that from an outward perspective its creates a "false positive" for people unfamiliar with the project. The couchbase-net-client project has this problem and we don't want to repeat that mistake...

- UseIndex_SelectDocuments also fails with a missing index:  GSI index beer_primary not found. Although the primary index does exist, but the test expects a _named primary index_ of "beer-primary". My thoughts are the same as above, however you had a note on the tests above indicating that the index needed to be created - you should probably add the same for clarity.

- Hmm, I mixed on the N1QlIndexType using strings. Enums are the convention and enforce type safety, but using strings is more flexible. TBH, it would be possible to just add a conversion layer as well. I would be willing to accept either one I think.

-Jeff

brantburnett:
@jeffrymorris 

Happy 4th to you, too!

I've updated all of the tests you mention to create the indexes as needed if they don't exist.  I also added the primary index to the global setup as well.  Now all of the tests should past, unless for some reason the bucket is not filled with the sample data.

I also gave the enumeration approach some more thought, and decided that reversing my earlier approach and using enumerations was probably better.  It's far more intuitive and easier to consume.  And it will be a very negligible amount of effort to add additional enumeration values in the future.

This pull should be good to go now, I hope!

Thanks,
Brant

jeffrymorris:
@brantburnett Thanks!

Issue #145, opened 2016-07-02
string.StartsWith, EndsWith


rogeralsing:
Any plans to include string.StartsWith or string.EndsWith ?
Currently we have to emulate this using `string.Substring(0,x.Length) == x` which doesnt read very well

brantburnett:
@rogeralsing 

Your right, that's an oversight.  Also, now that I look at it regular Contains only supports passing constants, not dynamic expressions.  Let me see what I can do with this.

brantburnett:
@rogeralsing 

See pull #146, that should give you the functionality you're looking for in the next release.  In the meantime, you can pull that down and build it locally if you need support sooner.

Thanks for the feedback.

Brant

PR #146, opened 2016-07-02
Issue #145: Support for StartsWith and EndWith

brantburnett:
Motivation
----------
Currently, only string.Contains is supported for string partial matching.
StartsWith and EndsWith support will make more advanced partial match
support easier.  Additionally, string.Contains will currently only work
with string constants, support for any Expression type will be more
useful.

Modifications
-------------
Added support for StartsWith and EndsWith to ContainsMethodCallTranslator.

Added support for dynamically building the right side of the LIKE
expression in N1QL if the input isn't a string constant.

Added unit and integration tests for the new functionality.

Results
-------
string.StartsWith and string.EndsWith are now supported in LINQ queries to
Couchbase.

Both of these methods, plus string.Contains, will also accept any
Expression type now, not just string ConstantExpressions.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
@@ -153,6 +153,45 @@ namespace Couchbase.Linq.IntegrationTests
         }
 
         [Test]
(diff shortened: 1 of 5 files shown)


jeffrymorris:
@brantburnett thanks!

Issue #153, opened 2016-07-22
Generated Queries Don't Allow The Use Of Array Indexes


brantburnett:
https://forums.couchbase.com/t/array-indexes-and-linq/9043/11

The extent name used in the ARRAY clause when making the index must match the name used in the generated query.  Currrently, extent names are always sequentially numbered in the generated query.

brantburnett:
Resolved by pull #154

jeffrymorris:
@brantburnett faster than me!

PR #154, opened 2016-07-22
Issue #153: Support for controlling array subquery extent names

brantburnett:
Motivation
----------
To use the new array index feature in Couchbase Server 4.5, the extent
names in array subqueries must exactly match the extent names used when
creating the array index.  This is incompatible with the current approach
of automatically generated extent names.

Modifications
-------------
Add IEnumerable<T> extension AsQueryable(extentName).  This overloads the
existing .Net AsQueryable() method.

When found, this expression node causes the ExtentNameProvider to use the
hard-coded extent name instead of an automatically generated name.

Created unit and integration tests for the new functionality.

Results
-------
It is now possible for SDK consumers to programatically control the extent
name used for array subqueries.  The consumer is responsible for ensuring
that they don't create naming conflicts in the resulting N1QL query.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
@@ -412,6 +412,35 @@ namespace Couchbase.Linq.IntegrationTests
         }
 
         [Test]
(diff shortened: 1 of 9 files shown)


jeffrymorris:
Thanks @brantburnett

Issue #157, opened 2016-09-26
Exception On Couchbase.Linq.BucketContext.GetDocumentId


Lightfall:
When KeyAttribute of Class is long GetId method throws exception when called by Save.

public class Product : IEntity
    {
        [Key]
        public long productId { get; set; }
....
}

**Message : Unable to cast object of type 'System.Int64' to type 'System.String'.
StackTrace:    at Couchbase.Linq.BucketContext.GetDocumentId[T](T document)
   at Couchbase.Linq.BucketContext.Save[T](T document)
   at Project.DataManager.SaveAll[T](Expression`1 property, IEnumerable`1 records, Exception& ex)**

Sorry for my bad english.

brantburnett:
@Lightfall 

This is currently a limitation of the key implementation, the property is required to be of type `string`.  As a workaround, I'd recommend adding the following property to your model:

```c#
[Key]
[JsonIgnore]
public string key { get { return productId.ToString(); } } 
```

I will leave this issue open so we can improve this in the future.

Thanks,
Brant

brantburnett:
@Lightfall 

This issue should be resolved now in release 1.2.1.  Thanks for bringing it to our attention.

Brant

PR #163, opened 2016-10-09
Issue #157: Allow non-string Key attributes on documents

brantburnett:
Motivation
----------
Currently, attributes marked as Key must be strings.  Converting other
types such as integers and GUIDs to strings would be simpler and more
obvious.

Modifications
-------------
Call ToString() instead of casting the result to a string.

To prevent a possible NRE, check for nulls before calling ToString() and
throw KeyMissingException.

Results
-------
Non-string key attributes are now supported.


jeffrymorris:
Thanks @brantburnett !

PR #165, opened 2016-11-07
Perform consistency checks when calling SubmitChanges

brantburnett:
Motivation
----------
It is generally expected behavior from an SDK like this to prevent
modifications to documents that have already been modified by other
processes, unless expressly overridden.

Modifications
-------------
Added SaveOptions class and optional parameter to SubmitChanges to allow
consistency checking to be disabled.  Defaults to enabled.

Updated SubmitChanges logic to pass the CAS from DocumentMetadata if
consistency checks are enabled.

Created CouchbaseConsistencyException, inherited from
CouchbaseWriteException, to differentiate between consistency failures and
other types of write failures.

Created unit and integration tests to cover the new functionality.

Results
-------
Consistency checking is now the default behavior for change tracking.  It
behaves in a simple and easy to understand manner for the consumer.

### DIFF
--- a/Src/Couchbase.Linq/SaveOptions.cs
+++ b/Src/Couchbase.Linq/SaveOptions.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Couchbase.Linq
+{
+    /// <summary>
+    /// Provides options controlling how changes are saved via <see cref="IBucketContext"/>.
+    /// </summary>
+    public class SaveOptions
+    {
+        /// <summary>
+        /// If true, before saving documents the Couchbase bucket is checked to ensure that another process
+        /// has not modified the document since it was read.  If it was modified, a
+        /// <see cref="CouchbaseConsistencyException"/> is thrown.  For new documents, the exception will be
+        /// thrown if the document already exists.  Defaults to true.
+        /// </summary>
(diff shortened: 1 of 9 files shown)


Src/Couchbase.Linq/SaveOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Couchbase.Linq
+{
+    /// <summary>
+    /// Provides options controlling how changes are saved via <see cref="IBucketContext"/>.
+    /// </summary>
+    public class SaveOptions
```
jeffrymorris: What if in the future, we wanted to have levels of consistency we wished to apply. For example, failing hard vs a warning if a consistency check failed....just an example. Are we future-proofing enough here?

jeffrymorris:
Thanks @brantburnett

Src/Couchbase.Linq/SaveOptions.cs
brantburnett: @jeffrymorris 

You make a good point.  Part of the problem is that we don't really have a good way to return such a status back to the user, we're taking the more LINQ like approach of throwing exceptions.

The best option I can think of off the top of my head would be:
1.  Make the setting an enumeration with three possible behaviors (ignore, fail and continue, fail and stop)
2.  if fail and stop, throw an immediate exception
3.  if fail and continue, collect the exception, and throw an AggregateException at the end with all collected errors

Thoughts?

Brant

PR #166, opened 2016-11-09
Improved document key selection for change tracking proxies

brantburnett:
Motivation
----------
Document key is inconsistently selected for change tracking proxies,
depending on alphabetical order and if properties are marked as virtual.

Modifications
-------------
For proxies, always use the document metadata when available, in lieu of
the key property on the POCO.  Do this using a simple typecast, rather
than reflection.

Also fixed problem where cache was never used except for proxies, not
for regular documents.

Results
-------
Consistent and predictable behavior for key selection on document
proxies during change tracking.


jeffrymorris:
Awesome work, @brantburnett !

PR #167, opened 2016-11-11
Configure AppVeyor CI

MikeGoldsmith:
## MOTIVATION
When trying to connect the Linq2Couchbase to project to the [AppVeyor CI](https://www.appveyor.com/) tool I found there is a local copy of .nuget (`src/.nuget`). This is an old version, 2.12, where 3+ is now available. Also, I don't know of the reason behind why there is a local copy. Most build tools, either through IDE or on command line, have their own nuget installs that they manage and update as appropriate.

## MODIFICATIONS
Remove the .nuget files and references in the .sln

## RESULT
The project now builds in AppVeyor and we don't have to maintain our own version of nuget that is packaged within the repository.

@brantburnett @jeffrymorris


jeffrymorris:
@MikeGoldsmith -

Should CI checks being failing or do I need to merge first?

-Jeff

MikeGoldsmith:
Sorry - I should have left a message. This isn't ready yet. There have been a few issues regarding nuget restore through AppVeyor. I'll add a label for now to show it's not ready for merge.

@jeffrymorris

MikeGoldsmith:
@jeffrymorris @brantburnett 
I think this is now ready for review. I'm not sure why there are two CI integrations (pr / branch) but it's building all projects and running the tests in `Couchbase.Linq.UnitTests`.

I'm going to look into how feasible doing unit tests in this CI context are and NetCore too. Currently the Nunit runner is 3.4 which I don't think is compatible - there is probably a way to patch that in but we can do that in another PR.

brantburnett:
@MikeGoldsmith @jeffrymorris 

It may be triggering two builds because there is both a global.json and a .sln file for the project.  The global.json is mainly there because project references in net core projects don't seem to work without it.  But just the .sln file should build everything.  I wonder if we could limit that somehow to just build one or the other.

Brant

jeffrymorris:
Thanks @MikeGoldsmith !

Issue #171, opened 2016-12-16
SelectMany unable to  unnest an array of strings


calhamdower:
Not Sure if there's meant to be support for SelectMany? Its not really detailed in the docs here that I could find.

Db.Query<Cat>().SelectMany(a => a.Feet).Explain();

I found though even running the Explain N1QL in CBQ wasnt a proper response:
"SELECT `Extent2`.* FROM `Testing` as `Extent1` INNER UNNEST `Extent1`.`teamList` as `Extent2` WHERE (`Extent1`.`type` = 'Sport')"
results in:
[
  {},
  {},
  {},
  {}
]
Maybe it should result in "SELECT `Extent2` as `result` FR......"

Couchbase.Linq.CouchbaseQueryException occurred
  HResult=-2146232832
  Message=An error occurred executing the N1QL query.  See the inner exception for details.
  Source=Couchbase.Linq
   InnerException: 
       HResult=-2146233088
       LineNumber=8
       LinePosition=9
       Message=Unexpected character encountered while parsing value: {. Path 'results', line 8, position 9.
       Path=results
       Source=Newtonsoft.Json

brantburnett:
@calhamdower 

Looking at it, I think the generated query looks correct.  But it's incorrectly trying to unwrap it from a results object.

Can you do a couple of things that will help me diagnose.

1. Can you provide a couple of sample documents and your POCOs?
2. Can you try running the query this way to see if it has different behavior?  It might build the lambdas differently:

```csharp
from p in Db.Query()
from q in p.Feet
select q
```

Thanks,
Brant

calhamdower:
@brantburnett 
Yes I tried that one earlier thinking it might resolve differently... though both:
var ids = (from p in Db.Query<Sport>()
                from q in p.TeamList
                select q).toList();
var ids2 =DAL.CbDal.BucketContext.Query<Sport>().Where(a=>a.Type=="Sport").SelectMany(a =>TeamList).ToList();
Result in the exact same fault. 

Fault occurs when the linq expression is evaluated via toList etc... so I think you're right in that it can't unwrap the result. when I run the N1QL expression in cbq however the result doesn't look as I'd expect also... I'd expect the result to appear as: 
```
"result": [
      "Team::9fb8b28813ea4132b8e8ce4f86c1664c",
      "Team::6012cf5cb6c84830a1a162157c1d3ad8"
    ]
rather than 
[
{},
{},
{},
{}
]
```
[...]

PR #175, opened 2016-12-20
Issue #171: Fix direct selection of unnest of scalar arrays

brantburnett:
Motivation
----------
Queries fail if you use a second from clause or .SelectMany() to unnest
an array of scalars and use that array directly as your select
projection.

Modifications
-------------
Test for arrays of scalars in select projections, and wrap them in
the result property of an object.  This accounts for the fact that N1QL
can't directly return an array of scalars, only objects.

Refactor the GetSelectClause method to separate
QuerySourceReferenceExpression select clauses for better code
organization and readability.

Add unit and integration tests for this type of unnest clause.  Added
overload of CreateN1QlQuery that returns the ScalarResultBehavior so we
can confirm that queries have ResultExtractionRequired = true.

Results
-------
These queries now function, existing behavior for other queries remains
unaffected.

The current definition of scalar arrays might be a bit broad at the
moment, detecting any .Net value types as well as strings.  This should
capture all numerics, enumerations, strings, and dates.  However, there
is a risk of unintentionally capturing structs.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
@@ -846,6 +846,25 @@ namespace Couchbase.Linq.IntegrationTests
         }
 
         [Test]
+        public void NestTests_Unnest_Scalar()
+        {
+            var bucket = ClusterHelper.GetBucket("beer-sample");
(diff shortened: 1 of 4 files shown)


jeffrymorris:
Thanks @brantburnett

PR #178, opened 2017-02-19
Upgrade to Couchbase SDK 2.4.0

brantburnett:
Motivation
----------
Have a consistent version for the .Net Full and .Net Standard versions
of Linq2Couchbase, with a consistent approach to logging.

Modifications
-------------
Upgraded all assemblies to use SDK 2.4.0.  Updated .Net Standard
libraries to use 1.6.1 of  NETStandard.Library.

Removed Common.Logging reference from .Net Standard assemblies.  Changed
all logging to use ILog instances from Couchbase.Logging.LogManager
instead of Common.Logging.LogManager.  These are per-instance of each
class instead of static.

Results
-------
All libraries are updated.  Logging still functions the same way for
.Net Full, but uses Microsoft logging for .Net Core.  The ILog instances
are no longer static in case the Couchbase LogManager is changed after
use to another logger.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests.NetCore/project.json
+++ b/Src/Couchbase.Linq.IntegrationTests.NetCore/project.json
@@ -2,7 +2,7 @@
   "version": "1.0.0-*",
 
   "dependencies": {
-    "NETStandard.Library": "1.6.0",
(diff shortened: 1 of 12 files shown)


jeffrymorris:
Thanks Brant!

PR #179, opened 2017-02-22
Use N1QL result streaming for query requests

brantburnett:
Motivation
----------
Reduce memory footprint for deserialized POCOs when executing large
queries.

Modifications
-------------
Turn on streaming by default for all queries except when change tracking
is enabled.

Add .UseStreaming(bool) extension to allow consumers to disable
streaming if desired.

Results
-------
Reduced memory consumption for large queries so long as they are
enumerated and not put into a list (.ToList()).

Note that this is specifically disabled if change tracking is enabled.
It is incompatible because of the streaming client doesn't use a data
mapper.  Additionally, it wouldn't be advantageous because the change
tracking system keeps the deserialized objects in memory anyway.


jeffrymorris:
Thanks Brant!

PR #180, opened 2017-03-05
Make Save behavior on BucketContext more logical during change tracking

brantburnett:
Motivation
----------
When using Save on BucketContext with documents other than new documents
it doesn't behave in a logical manner.

Modifications
-------------
Move RegisterChangeTracking call into Track method on BucketContext for
improved consistency.  Also ensure that Untrack calls
UnregisterChangeTracking.

When a document is removed then readded to the context, treat as a
modification instead of as a delete.

Calling Save on a document already flagged as modified has no effect.

Calling Save on an already tracked document flags the document as
modified and dirty so that it posts to Couchbase on SubmitChanges.

Results
-------
Behavior of Save is more consistent with what a consumer would expect.

### DIFF
--- a/Src/Couchbase.Linq.UnitTests/BucketContextTests.cs
+++ b/Src/Couchbase.Linq.UnitTests/BucketContextTests.cs
@@ -386,6 +386,105 @@ namespace Couchbase.Linq.UnitTests
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


jeffrymorris:
Thanks Brant!

PR #184, opened 2017-05-11
Fixed sporadic "Document does not contain valid MutationToken." error when saving documents

brendon1982:


brantburnett:
@brendon1982 

The fix itself looks good to me.  Thanks for the pull.

But I'm curious, why did you also update Json.net?  Generally speaking, we like to keep the version lower to increase compatibility with consumers who might be using older versions.

Brant

MikeGoldsmith:
@brendon1982 Thanks for the pull request!

The change looks to me good apart from a couple of things:
1) Upgrading JSON.NET, as Brant says, we try to keep it in line with the official Couchbase client
2) I don't think we need to update the project file to change the package hint paths, VS standard is to use `..\`

Thanks

jeffrymorris:
Hi @brendon1982 -

We would like to merge your pull request, but we need you to fix the issues brought up by @brantburnett and @MikeGoldsmith .

Also, the CI is failing so we'll need to resolve that as well. If you can't do this for any reason, let us know so we can take further action.

Thanks,

Jeff

brendon1982:
Hi @jeffrymorris 

Sorry for the delay. I completely agree with all the points you've raised. We originally forked the project and included the source into our project via a git sub module, we made the changes we did to accommodate this. I'll make another merge request with out the offending changes :D

Thanks

PR #185, opened 2017-05-19
Mutation State Fix

brendon1982:
Updated guard clause on BucketContext.AddToMutationState to match the logic in Couchbase .NET Client library, this was to fix an issue where sometimes the .NET Client Library returned an invalid Token.


brendon1982:
So I see the build is still failing, but I'm failing to see how the files I've changed has contributed to it's failure. I did make a new branch with only the change you guys wanted, however it seems as though the build is not fetching from my branch, it seems like it is fetching from master, which has the changes we don't want and the issue that is causing the build to fail.

I think I need to do a fresh fork and make the changes there because I don't want to disturb the current fork as our production project is currently using it.

PR #190, opened 2017-05-30
Remove ILRepack dependency and update to SDK 2.4.5

jeffrymorris:
Motivation
----------
We no longer need ILRepack and it doesn't support .NET Core, so removing
the dependency.

Modifications
-------------
 - Remove ILRepack package and project dependencies
 - Update to Couchbase.NET SDK 2.4.5
 - Revert Castle dependency to 3.3.3 (from 4.0.0)

Results
-------
Dependencies are now managed through NuGet.

### DIFF
--- a/Src/Couchbase.Linq/Couchbase.Linq.nuspec
+++ b/Src/Couchbase.Linq/Couchbase.Linq.nuspec
@@ -2,7 +2,7 @@
 <package >
   <metadata>
     <id>Linq2Couchbase</id>
-    <version>1.3.0</version>
+    <version>1.3.1</version>
     <title>The official Linq provider for Couchbase N1QL</title>
     <authors>Couchbase, Inc.</authors>
     <owners>Couchbase, Inc.</owners>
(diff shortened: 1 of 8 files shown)


Src/Couchbase.Linq/Couchbase.Linq.nuspec
```diff
@@ -15,11 +15,14 @@
...
       <group targetFramework="net45">
+        <dependency id="CouchbaseNetClient" version="2.4.5" />
         <dependency id="Newtonsoft.Json" version="9.0.1" />
         <dependency id="Common.Logging" version="3.3.1" />
+        <dependency id="Castle.Core" version="3.3.3" />
+        <dependency id="Remotion.Linq" version="2.1.1" />
       </group>
       <group targetFramework="netstandard1.5">
-        <dependency id="Castle.Core" version="4.0.0" />
+        <dependency id="Castle.Core" version="3.3.3" />
```
brantburnett: @jeffrymorris 

The minimum version for the .NET Standard 1.5 version really is 4.0.0 because that's the first version of Castle.Core that supports .NET Standard.  Version 1.3.1 now won't install on .NET Core applications.  I'll file a new pull to fix this back.

Brant

Issue #191, opened 2017-05-31
ExecuteAsync Can't Be Used Against Mock In-Memory Collections


brantburnett:
When performing unit tests using mock in-memory collections, ExecuteAsync fails with "ExecuteAsync is only supported on Couchbase LINQ queries."  We should probably include an exception for in-memory collections, and just execute synchronously and use Task.FromResult to return the result.

PR #194, opened 2017-05-31
Issue #191: Allow ExecuteAsync to be used with in-memory collections

brantburnett:
Motivation
----------
If working with a mock IBucketContext, calling Query<T> will normally
return an in-memory collection converted to an IQueryable.  In this
case, it will be useful if ExecuteAsync simply processes the in-memory
collection synchronously for the test rather than throwing an exception.

Modifications
-------------
Detect an in-memory collection (based on EnumerableQuery) and execute
synchronously.

Results
-------
Consumers mocking IBucketContext may now use ExecuteAsync against
in-memory collections.


jeffrymorris:
Thanks @brantburnett !

PR #196, opened 2017-05-31
Add support for Ordinal string comparison in linq expressions

drGarbinsky:
The ASP .NET OData query string parser uses the 3 argument overload of `String.Compare` to specify an ordinal string comparison. I don't thing N1QL is culture aware which is where this seems to be more relevant so it is probably just necessary to accommodate that overload in `StringComparisonExpressionTransformer` However adding support for this overload does tend to imply that the Ignore Case variations of the `StringComparison` options is supported so that may need a treatment of some sort.

### DIFF
--- a/Src/Couchbase.Linq.UnitTests/QueryGeneration/StringTests.cs
+++ b/Src/Couchbase.Linq.UnitTests/QueryGeneration/StringTests.cs
@@ -536,6 +536,95 @@ namespace Couchbase.Linq.UnitTests.QueryGeneration
         }
 
         [Test]
+        public void Test_StringCompare_EqualOrdinal()
+        {
+            var mockBucket = new Mock<IBucket>();
+            mockBucket.SetupGet(e => e.Name).Returns("default");
+
+            var query = from contact in QueryFactory.Queryable<Contact>(mockBucket.Object)
+                        where string.Compare(contact.FirstName, "M", StringComparison.Ordinal) == 0
+                        select new { contact.FirstName };
+
+            const string expected =
+                "SELECT `Extent1`.`fname` as `FirstName` FROM `default` as `Extent1` " +
+                "WHERE (`Extent1`.`fname` = 'M')";
+
+            var n1QlQuery = CreateN1QlQuery(mockBucket.Object, query.Expression);
+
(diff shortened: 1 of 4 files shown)


drGarbinsky:
PR adds support for `Ordinal`, `OrdinalIgnoreCase`, `CurrentCultureIgnoreCase`.  The `InvariantCulture` and `InvariantCultureIgnoreCase` options are not supported in .Net Core so I am unsure how to handle those.

brantburnett:
@drGarbinsky 

Your work on handling the comparison looks good.

Based on my research, N1QL does all string comparisons as ordinals.  So I think we should probably only support the Ordinal comparison options for compatibility.  However, would this cause a problem for OData compatibility?

Also, there seem to be some extra commits here that got picked up from elsewhere.  Could you rebase to master and squash to a single commit?

Brant

drGarbinsky:
I don't think that will cause a compatibility issue with OData as I am not aware of any part of that standard that provides for control over that behavior.

For the unsupported string comparison options should it throw a `NotSupportedExceptions` or just treat them as a normal sort?

-Can do on the rebase.

brantburnett:
@drGarbinsky 

I'd lean towards throwing a NotSupportedException for now.  Otherwise, we could run into backward compatibility issues if we tried to change it in the future.  This way we'd only be adding support for other values and not altering existing behavior if we add support for them in the future.

Brant

drGarbinsky:
@brantburnett  I've rebased and squashed.  I also added the exception for the unsupported comparison types.

jeffrymorris:
@drGarbinsky -

Thanks for the submission! It looks good, but before I merge, can you reformat your commit message in the following format:

Motivation
------------
The purpose of the commit.

Modifications
--------------
Description of what changed

Result
------------------
What a user should expect from the PR

For an example, you can see what @brantburnett did for [PR 194](https://github.com/couchbaselabs/Linq2Couchbase/pull/194).

Thanks!

Jeff

drGarbinsky:
@jeffrymorris  I think I have the commit message sorted out.

jeffrymorris:
Thanks @drGarbinsky

Issue #199, opened 2017-06-03
Left join not working when using DocumentTypeFilter


timtim002hk:
Left join not working, when i using DocumentTypeFilter. I would like to know how to fix this problem without remove DocumentTypeFilter.

**My Linq**:
var query = from s in context.Query<User>()
        join c in context.Query<Country>() on s.CountryKey equals Couchbase.Linq.N1QlFunctions.Key(c) into cg
        from c in cg.DefaultIfEmpty()
        select new { User = s, Country = c };

**Generated Statement**
SELECT `Extent1` as `User`, `Extent2` as `Country` FROM `courage` as `Extent1` LEFT JOIN `courage` as `Extent2` ON KEYS `Extent1`.`countryKey` WHERE (`Extent1`.`type` = 'user') AND (`Extent2`.`type` = 'country')

**Correct Statement**
SELECT `Extent1` as `User`, `Extent2` as `Country` FROM `courage` as `Extent1` LEFT JOIN `courage` as `Extent2` ON KEYS `Extent1`.`countryKey` WHERE (`Extent1`.`type` = 'user')

brantburnett:
@timtim002hk 

Based on your question, I'm guessing the problem is that the Extent2.type = 'country' filter is being applied after the join rather than before the join.  This is causing it to behave like an inner join instead of like a left join?

If so, that is, unfortunately, a known limitation currently: https://github.com/brantburnett/Linq2Couchbase/blob/master/Src/Couchbase.Linq/QueryGeneration/N1QLQueryModelVisitor.cs#L909

First of all, having an `Extent2.type = 'country'` clause is accurate, based on the LINQ statement being processed.  The presence of the DocumentTypeFilter means that some kind of filter is requested and should be present.  The issue is that it isn't being applied correctly.

Basically, I'm not certain of a way to actually replicate the desired behavior with N1QL.  It's a limitation of the ON KEYS clause for LEFT JOINs that you can't apply additional filtering before the join is processed.  But this filter is something included in the LINQ statement, so the query generator is making a best effort to include it in the query (in an incorrect way in this case).

A few ideas I've considered:
1. Ignore filter clauses before the join (seems like poor form)
2. Provide an overload of the Query method with an option to suppress filters which you can use if you're doing joins
3. Find some way to provide the behavior described in LINQ using a construct other than a JOIN (may not be possible or may be very messy)
[...]

PR #200, opened 2017-06-03
Issue #199: Add option to suppress document filters

brantburnett:
Motivation
----------
When performing LEFT JOINs, filters which are automatically injected via
the DocumentFilterManager can produce unexpected behaviors.  This is due
to a N1QL limitation that you cannot apply filters before a JOIN, only
after the JOIN is complete.

Modifications
-------------
Added BucketContext.Query overload that allows the user to suppress the
filters, such as the DocumentTypeFilter.

Added unit tests for all variants to ensure proper behavior.

Results
-------
When performing a LEFT JOIN, the user can apply
BucketQueryOptions.SuppressFilters to the right hand side of the join.
This will prevent the LINQ expressions which cannot be translated to
N1QL from being applied to the query.


jeffrymorris:
Thanks @brantburnett

Issue #203, opened 2017-07-17
Add Support for comparison of DateTimeOffset against a string constant


drGarbinsky:
Repro:

1) Create a model that has a DateTimeOffset for a property.
2) Create a LINQ statement that uses that property in the where clause: MyProp >= "2016-12-29T09:14:32.8445409Z"

Result: the constant appears in the generated N1qL statement without surrounding quotes.

drGarbinsky:
PR in-bound

PR #204, opened 2017-07-17
Issue #203 Add Support for comparison of DateTimeOffset against a str…

drGarbinsky:
Issue #203 Add Support for comparison of DateTimeOffset against a string constant 

Motivation
----------
Comparing DateTimeOffset values to a constant string value in a where clause will result in an unquoted string.

Modifications
-------------
-Add an ORed expression to an if statement in N1QlExpressionTreeVisitor.VisitConstant

Results
-------
LINQ Where clause can properly quote constants when compared to DateTimeOffset properties.

### DIFF
--- a/Src/Couchbase.Linq.NetStandard/Couchbase.Linq.NetStandard.csproj
+++ b/Src/Couchbase.Linq.NetStandard/Couchbase.Linq.NetStandard.csproj
@@ -386,6 +386,9 @@
     <Compile Include="..\Couchbase.Linq\Versioning\VersionProvider.cs">
       <Link>Versioning\VersionProvider.cs</Link>
     </Compile>
+    <Compile Include="..\Couchbase.Linq\UnixMillisecondsDateTimeOffset.cs">
+      <Link>UnixMillisecondsDateTimeOffset.cs</Link>
+    </Compile>
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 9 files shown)


MikeGoldsmith:
Hi @drGarbinsky 
We actually saw and discussed your work not long before you created the PR. I think this is a good improvement.
I also don't think the AppVeyor failure is down to this change as it looks like it failed to download file before building. I'll look more into that.

Thanks

drGarbinsky:
I'll take a crack at it today. The recommended change makes sense.

drGarbinsky:
Updated the PR, Let me know if I approached the change correctly. Added a DateTimeOffset property for Beer which now diverges from the sample data in terms of schema. Not sure if that's the correct approach.

Src/Couchbase.Linq.NetStandard/UnixMillisecondsDateTimeOffset.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
```
brantburnett: Instead of repeating this class for NetStandard, it should just be added as a Link to the file in the other project.  This must be done by manually editing the csproj file in a text editor.

drGarbinsky:
File Linked

brantburnett:
@MikeGoldsmith 

This looks good to me if you want to merge it.

Brant

PR #208, opened 2017-09-09
Fix startup exception if framework doesn't have TrimStart/TrimEnd

brantburnett:
Motivation
----------
All LINQ queries will fail with a type initializer exception if the
target .NET Framework doesn't implement string.TrimStart and
string.TrimEnd with no parameters.

Modifications
-------------
Load both TrimStart and TrimEnd with no parameters and with a char[]
parameter.

Skip null MethodInfo values when building the
DefaultMethodCallTranslatorProvider, so unsupported methods on the given
framework are ignored.

Results
-------
Queries now function correctly.


jeffrymorris:
Thanks @brantburnett

PR #210, opened 2017-09-13
Change to Couchbase SDK cluster versioning

brantburnett:
Motivation
----------
Current cluster versioning is incompatible with Couchbase Server 5.0 due
to RBAC.  Also, the current approach doesn't handle clusters with mixed
versions across nodes.

Modifications
-------------
Upgrade to Couchbase SDK 2.5.0.

Replaced the internal cluster version approach with the new Couchbase
SDK approach.

Updated unit tests to handle the new cluster version source.

Fixed one integration test that should expect IndexScan plan for pre-5.0
and IndexScan2 for 5.0.

Results
-------
Cluster version is more accurate, and also includes build numbers which
may be useful for some features in the future.

Cluster version is provided correctly for Couchbase Server 5.0 clusters.


jeffrymorris:
Thanks @brantburnett

PR #214, opened 2017-09-17
Disable result streaming by default

brantburnett:
Motivation
----------
Streaming of query results is incompatible with custom deserializers, as
internally it uses Newtonsoft.Json.  Therefore, enabling streaming by
default was a breaking change for consumers using custom serializers.

Modifications
-------------
Disable query streaming by default.

Added streaming documentation, which was missing.

Results
-------
Existing consumers who have not manually specified a streaming mode will
now not use streaming.  Consumers performing change tracking will be
unaffected, as this mode already disabled streaming.  Compatibility with
custom serializers is now the default mode.

There is a small risk of an unexpected increase in memory consumption
upon upgrade if users were expecting the default setting to be enabled
and are querying very large result sets which are processed iteratively.
This should be addressed in release notes.

### DIFF
--- a/Src/Couchbase.Linq/Execution/IBucketQueryExecutor.cs
+++ b/Src/Couchbase.Linq/Execution/IBucketQueryExecutor.cs
@@ -28,7 +28,7 @@ namespace Couchbase.Linq.Execution
         /// <summary>
         /// Specifies if the query results should be streamed, reducing memory utilzation for large result sets.
         /// </summary>
-        /// <remarks>The default is true.</remarks>
+        /// <remarks>The default is false.</remarks>
         bool UseStreaming { get; set; }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


Src/Couchbase.Linq/Execution/IBucketQueryExecutor.cs
```diff
@@ -28,7 +28,7 @@ internal interface IBucketQueryExecutor : IQueryExecutor
         /// <summary>
         /// Specifies if the query results should be streamed, reducing memory utilzation for large result sets.
         /// </summary>
-        /// <remarks>The default is true.</remarks>
+        /// <remarks>The default is false.</remarks>
         bool UseStreaming { get; set; }
```
jeffrymorris: Potentially breaking change; imagine a somebody has a large query that previously worked and now they get an OOM.

brantburnett:
@jeffrymorris 

I thinking making the streaming parser pluggable seems like a good idea, but probably more long term.  And probably something to complete at the SDK level, as well.

For now, I think we're in between a bit of a rock and a hard place, because the previous release that turned on streaming by default was also a breaking change.  So whichever way we go is breaking.  Based on that, I think we should choose the option that's the best overall.  Between enabling and disabling streaming, I figure disabling makes the most sense.  For most queries streaming is overkill, so turning it on when needed feels the most natural.

jeffrymorris:
@brantburnett thanks!

PR #216, opened 2017-09-20
Use new SELECT RAW syntax to extract scalar results on Server 5.0

brantburnett:
Motivation
----------
Couchbase Server 5.0 offers a new SELECT RAW syntax which can extract
raw scalars in the result set, rather than the JSON objects always
returned by SELECT.  The current workaround is to use this syntax:

`SELECT Extent1.property as result FROM bucket as Extent1`

The `result` attribute is then extracted from the object for each row
returned.  The new syntax can be:

`SELECT RAW Extent1.property FROM bucket as Extent1`

This syntax doesn't require the wrapping object with a `result`
attribute.  This reduces data on the wire, heap allocations during
deserialization, and the number of operations to be performed processing
each result row.

This support will also be useful in future commits to implement array
subqueries, another new 5.0 feature.

Modifications
-------------
Use SELECT RAW wherever applicable instead of result extraction, so long
as the cluster is on version 5.0 or greater.

Test each of these circumstances to ensure property queries are
generated on pre-5.0 and post-5.0 servers.

Results
-------
The following types of queries now use SELECT RAW for performance if the
cluster version is 5.0 or greater:

- Selecting a document POCO directly without projection, so long as
change tracking is not enabled
- Selecting scalar properties directly
- Subqueries returning scalar properties directly
- Standalone aggregates

### DIFF
--- a/Src/Couchbase.Linq/Versioning/FeatureVersions.cs
+++ b/Src/Couchbase.Linq/Versioning/FeatureVersions.cs
@@ -27,5 +27,10 @@ namespace Couchbase.Linq.Versioning
         /// Version where support was added for array indexes.
         /// </summary>
         public static readonly ClusterVersion ArrayIndexes = new ClusterVersion(new Version(4, 5, 0));
+
+        /// <summary>
+        /// Version where support was added for SELECT RAW
+        /// </summary>
+        public static readonly ClusterVersion SelectRaw = new ClusterVersion(new Version(5, 0, 0));
     }
 }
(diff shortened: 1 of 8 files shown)


Src/Couchbase.Linq/Versioning/FeatureVersions.cs
```diff
@@ -27,5 +27,10 @@ internal static class FeatureVersions
         /// Version where support was added for array indexes.
         /// </summary>
         public static readonly ClusterVersion ArrayIndexes = new ClusterVersion(new Version(4, 5, 0));
+
+        /// <summary>
+        /// Version where support was added for SELECT RAW
```
MikeGoldsmith: added support for SELECT RAW

brantburnett: @MikeGoldsmith Not sure I follow what you're asking for here, doesn't look like a typo to me.

MikeGoldsmith: You are right, had a total memory meltdown. Description is fine 👍

jeffrymorris:
@brantburnett thanks!

Issue #219, opened 2017-11-06
_cachedKeyProperties in BucketContext class could be static to run faster !!??


lpyedge:
_cachedKeyProperties in BucketContext class could be static to run faster !!??

brantburnett:
@lpyedge 

Good catch, it was intended to be but it looks like it was overlooked.  Thanks for bringing it to our attention.

MikeGoldsmith:
Closing this issue as the fix to make CackedKeyProperties static has been merged.

PR #221, opened 2017-11-07
Issue #219: Make document key attribute cache static

brantburnett:
Motivation
----------
Reflected attributes cannot change at runtime, so performance can be
increased by caching them for the application lifetime.

Modifications
-------------
Change the attribute cache to a static private field, so it is shared
across instances of BucketContext.

Results
-------
Keys are now cached for the application lifetime.

### DIFF
--- a/Src/Couchbase.Linq/BucketContext.cs
+++ b/Src/Couchbase.Linq/BucketContext.cs
@@ -22,7 +22,7 @@ namespace Couchbase.Linq
     /// </summary>
     public class BucketContext : IBucketContext, IChangeTrackableContext
     {
-        private readonly ConcurrentDictionary<Type, PropertyInfo>_cachedKeyProperties = new ConcurrentDictionary<Type, PropertyInfo>();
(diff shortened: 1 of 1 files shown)


brantburnett:
@jeffrymorris 

ping for review

Src/Couchbase.Linq/BucketContext.cs
```diff
@@ -22,7 +22,7 @@ namespace Couchbase.Linq
     /// </summary>
     public class BucketContext : IBucketContext, IChangeTrackableContext
     {
-        private readonly ConcurrentDictionary<Type, PropertyInfo>_cachedKeyProperties = new ConcurrentDictionary<Type, PropertyInfo>();
+        private static readonly ConcurrentDictionary<Type, PropertyInfo> s_cachedKeyProperties = new ConcurrentDictionary<Type, PropertyInfo>();
```
MikeGoldsmith: Typically we follow ReSharper's code styling which would recommend a capitalised static variable name; 
 eg *CachedKeyProperties*

brantburnett:
@MikeGoldsmith good catch, fixed.

Issue #225, opened 2018-01-16
Support For Querying Dictionaries Serialized As JSON Objects


brantburnett:
Checklist:

- [x] Index accessor `dictionary["key"]` -> `dictionary.key`
- [x] Key presense `dictionary.ContainsKey["key"]` -> `dictionary.key IS NOT MISSING`
- [ ] Enumerate keys `dictionary.Keys` -> `OBJECT_NAMES(dictionary)`
- [ ] Enumerate values `dictionary.Values` => `OBJECT_VALUES(dictionary)`

brantburnett:
@dgrim-CE

Here is the issue related to your request.  Thanks!

PR #226, opened 2018-01-16
Issue #225: Support for dictionary index accessors

brantburnett:
Motivation
----------
.NET dictionarys are serialized as JSON objects, but there is currently
no way to query against the inner contents of these objects via LINQ.

Modifications
-------------
Adjusted DefaultMethodCallTranslatorProvider to support detecting
generic methods based on generic parameters of the type, not just the
method.

Added a method call translator for IDictionary.get_Item, with supporting
unit tests.

Results
-------
Calls to `dictionary["key"]` will now be correctly translated to
`dictionary.key` in N1QL.


jeffrymorris:
Thanks @brantburnett

PR #227, opened 2018-01-17
Issue #225: Support for Dictionary<TKey, TValue>.ContainsKey(TKey)

brantburnett:
Motivation
----------
Allow queries to test for the presense of a particular key on an object
which is serialized from a Dictionary implementation.

Modifications
-------------
Implement DictionaryContainsKeyMethodCallTranslator, along with matching
unit tests.

Fix a bug in the DefaultMethodCallTranslatorProvider related to
detecting interface method translations from a class.

Results
-------
ContainsKey method calls on dictionaries may now be used in queries.

### DIFF
--- a/Src/Couchbase.Linq/QueryGeneration/MethodCallTranslators/DictionaryContainsKeyMethodCallTranslator .cs
+++ b/Src/Couchbase.Linq/QueryGeneration/MethodCallTranslators/DictionaryContainsKeyMethodCallTranslator .cs
@@ -1,0 +1,41 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using System.Reflection;
+namespace Couchbase.Linq.QueryGeneration.MethodCallTranslators
+{
+    internal class DictionaryContainsKeyMethodCallTranslator  : IMethodCallTranslator
+    {
+        private static readonly MethodInfo[] SupportedMethodsStatic = {
+            typeof (IDictionary).GetMethod("Contains"),
+            typeof (IDictionary<,>).GetMethod("ContainsKey")
+        };
+
+        public IEnumerable<MethodInfo> SupportMethods => SupportedMethodsStatic;
+
+        public Expression Translate(MethodCallExpression methodCallExpression, N1QlExpressionTreeVisitor expressionTreeVisitor)
+        {
+            if (methodCallExpression == null)
+            {
+                throw new ArgumentNullException(nameof(methodCallExpression));
+            }
+
+            if (methodCallExpression.Arguments[0] is ConstantExpression keyExpression)
+            {
+                var expression = expressionTreeVisitor.Expression;
+
+                expressionTreeVisitor.Visit(methodCallExpression.Object);
+                expression.Append('.');
+                expression.Append(N1QlHelpers.EscapeIdentifier(keyExpression.Value.ToString()));
+                expression.Append(" IS NOT MISSING");
+            }
+            else
+            {
+                throw new NotSupportedException("Dictionary keys must be constants");
+            }
+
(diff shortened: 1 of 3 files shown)


Src/Couchbase.Linq/QueryGeneration/MethodCallTranslators/DictionaryContainsKeyMethodCallTranslator .cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Couchbase.Linq.QueryGeneration.MethodCallTranslators
+{
+    internal class DictionaryContainsKeyMethodCallTranslator  : IMethodCallTranslator
+    {
+        private static readonly MethodInfo[] SupportedMethodsStatic = {
+            typeof (IDictionary).GetMethod("Contains"),
+            typeof (IDictionary<,>).GetMethod("ContainsKey")
+        };
+
+        public IEnumerable<MethodInfo> SupportMethods => SupportedMethodsStatic;
```
MikeGoldsmith: The static and instance properties could be combined like this:

```
public static readonly IEnumerable<MethodInfo> SupportedMethods = new [] {
    typeof (IDictionary).GetMethod("Contains"),
    typeof (IDictionary<,>).GetMethod("ContainsKey")
};
```

Src/Couchbase.Linq/QueryGeneration/MethodCallTranslators/DictionaryContainsKeyMethodCallTranslator .cs
```diff
@@ -0,0 +1,41 @@
...
+            }
+
+            if (methodCallExpression.Arguments[0] is ConstantExpression keyExpression)
+            {
+                var expression = expressionTreeVisitor.Expression;
+
+                expressionTreeVisitor.Visit(methodCallExpression.Object);
+                expression.Append('.');
+                expression.Append(N1QlHelpers.EscapeIdentifier(keyExpression.Value.ToString()));
+                expression.Append(" IS NOT MISSING");
```
MikeGoldsmith: Minor thing: it may be worth creating a helper that contains constant values to help reduce duplication / slightly improve efficiency.

Src/Couchbase.Linq/QueryGeneration/MethodCallTranslators/DictionaryContainsKeyMethodCallTranslator .cs
brantburnett: Right now it's an instance method so it can be exposed on the interface, which allows us to get it without so much reflection: https://github.com/couchbaselabs/Linq2Couchbase/blob/master/Src/Couchbase.Linq/QueryGeneration/DefaultMethodCallTranslatorProvider.cs#L36

We could change it, but it would require a large refactor and should probably be a separate issue.

Src/Couchbase.Linq/QueryGeneration/MethodCallTranslators/DictionaryContainsKeyMethodCallTranslator .cs
brantburnett: Yeah, I've been thinking about that, too.  I'll file an issue for it so we can keep track.

Src/Couchbase.Linq/QueryGeneration/MethodCallTranslators/DictionaryContainsKeyMethodCallTranslator .cs
MikeGoldsmith: Sure, ok - makes sense.

brantburnett:
@jeffrymorris

jeffrymorris:
Thanks @brantburnett

Issue #229, opened 2018-01-24
Support For Query Cancellation


brantburnett:
It should be possible to extend ExecuteAsync to accept a CancellationToken.  If this is passed on to the N1QL execution, it should cancel the HTTP request and stop query execution.

PR #231, opened 2018-02-17
Issue #229: Support for async query cancellation

brantburnett:
Motivation
----------
Allow long running LINQ queries executing asynchronously to be
cancelled if the result is no longer required.  For example, if an
incoming HTTP request has its connection dropped before the query
completes.

Modifications
-------------
Added overloads to ExecuteAsync to accept a CancellationToken, defaulted
to none.

Updated backing methods to accept a CancellationToken and pass it on the
the Couchbase SDK.

Results
-------
Async queries may now be cancelled.


Src/Couchbase.Linq/Execution/IBucketQueryExecutor.cs
```diff
@@ -43,15 +44,17 @@ internal interface IBucketQueryExecutor : IQueryExecutor
         /// </summary>
         /// <typeparam name="T">Type returned by the query.</typeparam>
         /// <param name="queryRequest">Request to execute.</param>
+        /// <param name="cancellationToken">Cancellation token.</param>
         /// <returns>Task which contains a list of objects returned by the request when complete.</returns>
-        Task<IEnumerable<T>> ExecuteCollectionAsync<T>(LinqQueryRequest queryRequest);
+        Task<IEnumerable<T>> ExecuteCollectionAsync<T>(LinqQueryRequest queryRequest, CancellationToken cancellationToken);
```
MikeGoldsmith: Changing the interface definition here can cause semver problems with dependent projects. Instead, we've found it better to provide an overload with the additional parameters with implementations redirected to the new implementation with defaults.

Src/Couchbase.Linq/Extensions/QueryExtensions.cs
```diff
@@ -306,7 +307,22 @@ public static IQueryable<T> UseIndex<T>(this IQueryable<T> source, string indexN
         /// <example>
         /// var results = await query.ExecuteAsync();
```
MikeGoldsmith: If we're updating method summaries, may as well update this one too.

Src/Couchbase.Linq/Execution/BucketQueryExecutor.cs
```diff
@@ -257,9 +259,9 @@ public T ExecuteSingle<T>(QueryModel queryModel, bool returnDefaultWhenEmpty)
             return result;
         }
 
-        public async Task<T> ExecuteSingleAsync<T>(LinqQueryRequest queryRequest)
+        public async Task<T> ExecuteSingleAsync<T>(LinqQueryRequest queryRequest, CancellationToken cancellationToken)
         {
-            var resultSet = await ExecuteCollectionAsync<T>(queryRequest);
+            var resultSet = await ExecuteCollectionAsync<T>(queryRequest, cancellationToken);
```
MikeGoldsmith: I think this should this have a ConfigureAwait(false) too.

Src/Couchbase.Linq/Execution/IBucketQueryExecutor.cs
brantburnett: This is an internal interface, so I was able to change it freely.

Src/Couchbase.Linq/Extensions/QueryExtensions.cs
brantburnett: Updated to show the cancellationToken parameter on two methods.

Src/Couchbase.Linq/Execution/BucketQueryExecutor.cs
brantburnett: Done

Src/Couchbase.Linq/Execution/IBucketQueryExecutor.cs
MikeGoldsmith: Ah, missed it is internal as the file was truncated for review.

Issue #232, opened 2018-03-13
Support for ANSI joins in Couchbase Server 5.5


brantburnett:
https://blog.couchbase.com/ansi-join-support-n1ql/

- [x] Basic INNER JOIN
- [x] Basic LEFT JOIN
- [x] Basic INNER NEST
- [x] Basic LEFT OUTER NEST
- [x] USE INDEX hint on the right-hand extent
- [x] USE HASH hint
- [x] List of keys NEST
- [x] USE INDEX, USE HASH on NEST operations

jeffrymorris:
@brantburnett with the merging of #233 can this be closed?

brantburnett:
@jeffrymorris not quite, #233 was step one, I’ve got some more coming.  This issue is more like an epic.

brantburnett:
I've removed the following to-dos from the list:

- RIGHT OUTER JOIN - Not possible with LINQ, just reverse the order of the extents and get a LEFT OUTER JOIN.  This is what N1QL does behind the scenes anyway, so there is no special reason to support it.
- Join on IN array clause - There is no matching LINQ construct, joins in LINQ are always equi-joins

brantburnett:
Added "List of keys NEST" as a required feature because:

- The new LINQ syntax for ANSI NEST doesn't handle a list of keys due to LINQ limitations
- The LINQ ".Nest" and ".LeftOuterNest" extensions we add does this, but always uses the ON KEYS syntax in the output N1QL
- Per @miaobingjie The ON KEYS syntax cannot be used in combination with ANSI join syntax, so using the LINQ ".Nest" or ".LeftOuterNest" will cause failures if mixed with any other joins

PR #233, opened 2018-03-13
Issue #232: Support for simple ANSI joins

brantburnett:
Motivation
----------
Make use of the new ANSI join features in Couchbase 5.5.

Modifications
-------------
Break up N1QlFromQueryPart into separate classes depending on the type of
join being defined.

Always use ANSI joins for INNER JOIN and LEFT JOIN when Couchbase Server
5.5 is detected (see
https://community.couchbase.com/main/groups/14079/lounge/posts/162691)

Supporting unit and integration tests.

Results
-------
ANSI joins are used whenever they are available, including when
N1QlFunctions.Key is present in the equijoin.


jeffrymorris:
Awesome work @brantburnett! Thanks for the submission!

Issue #223, opened 2017-11-16
querying datetime fields (that is stored as long/unixtime)


alexdoan102:
Hi, 
 based on your https://github.com/couchbaselabs/Linq2Couchbase/blob/master/docs/date-handling.md
 documentation, to query datetime fields, you need to store dates as  ISO 8601.  I stored my dates as long/unixtime (for compatibility with java/android cblite projects which uses long/unixtime),  

 Is there a way to make queries work with datetime fields (where the datetime is stored as long/unixtime).
 I current use a jsonConverter to convert long/unixtime to datetime fields on serliaziation/deseralization.
This works fine, but now i cannot query datetime fields, since my datetime values are stored as "long" in the couchcbase database.

 I noticed there is a UnixMillescondsDateTime.cs  class in your sourscecode.  Is there a way to force datetime to long for querying?

brantburnett:
@alexdoan102 

Currently, this isn't supported, but I do think supporting it is a good idea.  While ISO8601 is one of the more prevalent storage methods, it certainly isn't the only one.

The key here is that the query generator needs some way to know how the column is stored in the database.  That way it can choose whether or not to wrap in STR_TO_MILLIS.  I've had a couple of thoughts about how this might be accomplished:

1.  An attribute that can be put on the property (has the advantage of simplicity)
2.  Information retrieved from the IExtendedTypeSerializer (more flexible and powerful, but more difficult to implement)
3.  A blend of both approaches, where the default serializer looks for the attribute but you could override with a more advanced behavior (the approach I'm leaning towards)
4.  Methods that could be included in the LINQ query to indicate how to handle it (I'm not a fan of this approach because it requires repeating this logic in every query)

I'd be interested in your thoughts, or if you have any other suggestions.

As a (slightly cumbersome) workaround, some users store both the Unix and ISO8601 formats in their documents.  Including the second format has the advantage of being more human readable for debugging, and you could use it for queries.

alexdoan102:
@brantburnett 

option #1 works best for my current issue.  (see attached screenshot)

option #2 and #3 is certainly fine, but certainly way more than what I need right now. (but future proofing is always a good idea)  ==> this is up to you.

option #4 is bad, I agree with you, it would require modifying all the queries and repeating logic.

[...]

PR #240, opened 2018-04-21
Issue #223: DateTime fields as Unix milliseconds

brantburnett:
Motivation
----------
Support date/time comparisons and functions on values persisted as
milliseconds since the Unix epoch.

Modifications
-------------
Upgrade to Couchbase SDK 2.6.0-beta.

For the default serializer, detect the presence of the
UnixMillisecondsConverter on an attribute to indicate that the attribute
is persisted as milliseconds.  For custom serializers, allow them to
implement IDateTimeSerializationFormatProvider to provide this
information.

When rendering UnixMilliseconds.ToDateTime (used internally for by
DateTimeComparisonExpressionTransformer), skip the STR_TO_MILLIS
conversion step for these attributes.

For DateTime parameters to N1QlFunctions methods, add a MILLIS_TO_STR
conversion for these attributes.

Refactor DocumentProxyDataMapper and BucketQueryExecutor so that they
can share a serializer.

Results
-------
Fields marked with UnixMillisecondsConverter are automatically supported
on queries for comparisons and date/time functions if using the default
serializer.  Custom serializers may implement
IDateTimeSerializationProvider to get the same support.

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/Documents/Beer.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/Documents/Beer.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.ComponentModel.DataAnnotations;
+using Couchbase.Core.Serialization;
 using Newtonsoft.Json;
 
 namespace Couchbase.Linq.IntegrationTests.Documents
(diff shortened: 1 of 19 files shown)


jeffrymorris:
Thanks @brantburnett

Issue #243, opened 2018-04-27
ANSI NEST extents can't be used with array subqueries


brantburnett:
Example query:

```cs
var query = from airline in context.Query<Airline>()
                join route in context.Query<Route>()
                    on airline.Iata equals route.Airline into routes
                select new { name = airline.Name, routes = routes.Where(p => p.DestinationAirport == "SCO").ToList() };
```

This query should result in the array being filtered in the select projection using a subquery.  However, it currently results in this NotSupportedException:
From Clauses May Not Reference Any Query Source Other Than The Grouping Subquery

Instead, the generated query should be something like this:

```sql
SELECT `Extent1`.`name` as `name`,
ARRAY `Extent3` FOR `Extent3` IN `Extent2` WHEN `Extent3`.`destinationAirport` = 'SCO' END as `routes`
FROM `travel-sample` as `Extent1`
LEFT OUTER NEST `travel-sample` as `Extent2`
ON (`Extent1`.`iata` = `Extent2`.`airline`) AND (`Extent2`.`type` = 'route')
WHERE (`Extent1`.`type` = 'airline')
```

PR #244, opened 2018-04-27
Issue #243: Allow ANSI NEST extents to be used in array subqueries

brantburnett:
Motivation
----------
The resulting extent from an ANSI NEST is an array, which should be
usable in subqueries to apply additional filtering or projections.

Modifications
-------------
When parsing a subquery from clause, recognize a group join as a NEST
and treat it as an array extent instead of throwing an error.

Results
-------
Additional filters and projections can now be applied to the output of
ANSI NEST operations.


jeffrymorris:
Thanks @brantburnett

Issue #254, opened 2018-09-08
Serialization Converter Isn't Applied When Attribute Is On A Class


brantburnett:
If `[JsonConverter(...)]` is applied to a type, and that type is stored in a property, the DefaultSerializationConverterProvider doesn't correctly detect it and apply the ISerializationConverter.  If the JsonConverter attribute is applied to the property instead it is correctly applied.

Reported by @dsenic, see https://forums.couchbase.com/t/linq2couchbase-generates-invalid-where-clause/18257 and https://github.com/dsenic/Linq2CoucbaseWhereQueryGeneration

PR #255, opened 2018-09-13
Issue #254: Implement more complete JsonConverter detection

brantburnett:
Motivation
----------
Current detection of JsonConverters is only finding converters applied
directly to properties, not those applied to classes or applied via
JsonSerializationSettings.Converters.

Modifications
-------------
Replicate the logic from Newtonsoft.Json to find the effective converter
(this logic is not currently exposed by Newtonsoft.Json for consumption,
that I can find).

Results
-------
Class JsonConverter attributes and JsonSerializationSettings.Converters
now apply.

We still need to work on implementing support for
[JsonProperty(ItemConverter=typeof(...))] attributes to handle items
within collection properties.


jeffrymorris:
Thanks @brantburnett

PR #280, opened 2020-02-11
Optional implementation to get the linq2couchbase log

WallaceValadao:
I see Couchbase SDK is going to change the log so i alter way to capture query log after transformation linq to n1ql. If this pull request be aprove, I will change the rest of the code.
The programer need only implemented the interface.

### DIFF
--- a/Src/Couchbase.Linq/BucketContext.cs
+++ b/Src/Couchbase.Linq/BucketContext.cs
@@ -42,6 +42,16 @@ namespace Couchbase.Linq
         }
 
         /// <summary>
+        /// Creates a new BucketContext for a given Couchbase bucket.
+        /// </summary>
+        /// <param name="bucket">Bucket referenced by the new BucketContext.</param>
+        /// <param name="Log">log capture interface</param>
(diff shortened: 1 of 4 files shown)


brantburnett:
@WallaceValadao This would be a breaking change for most of our users, who already receive the logs via the logging infrastructure built into the Couchbase SDK. I think any change we implement should ensure backward compatibility with the current logging approach.

Also, can I get a better understanding of what this change is trying to accomplish? If it's related to the SDK 3 logging changes, there are actually a lot more changes we'll be implementing for SDK 3 support. We should probably make a milestone for SDK 3 compatibility and file issues for all the things we want to address.

WallaceValadao:
Exactly @brantburnett, due to future changes this way will be easy to use new couchbase sdk and stay open to the programmer use other log in BucketContext. But I don't know if this is the best solution.

brantburnett:
@WallaceValadao 

Here's the issue I've filed to address logging in SDK 3.0. Please feel free to respond on it if you disagree with the approach.

https://github.com/couchbaselabs/Linq2Couchbase/issues/282

PR #285, opened 2020-04-15
Fix and update broken links

ofthelit:
One wasn't a proper markdown link and some others didn't work anymore.


brantburnett:
Thanks @ofthelit !

Issue #281, opened 2020-02-24
Rewrite Async Queries To Use IAsyncEnumerable


brantburnett:
- [x] Async enumeration
- [x] AsAsyncEnumerable
- [x] ToListAsync
- [x] FirstAsync
- [x] FirstOrDefaultAsync
- [x] SingleAsync
- [x] SingleOrDefaultAsync
- [x] AnyAsync
- [x] AllAsync
- [x] SumAsync
- [x] AverageAsync
- [ ] MinAsync
- [ ] MaxAsync
- [x] CountAsync
- [x] LongCountAsync

PR #314, opened 2020-06-13
Add support for CountAsync and LongCountAsync

brantburnett:
Motivation
----------
Support asynchronous queries which return counts.

Modifications
-------------
Create extension methods, an expression node, a result operator, and
streamed data classes to handle CountAsync and LongCountAsync.

Add tests for the async methods.

Results
-------
CountAsync or LongCountAsync may be called on IQueryable and they
function correctly so long as the IQueryable is from a
CollectionContext.

Relates to #281

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/CountQueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/CountQueryTests.cs
@@ -1,0 +1,89 @@
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Couchbase.Linq.Extensions;
+using Couchbase.Linq.IntegrationTests.Documents;
+using NUnit.Framework;
+
+namespace Couchbase.Linq.IntegrationTests
+{
+    [TestFixture]
+    public class CountQueryTests : N1QlTestBase
+    {
+        [Test]
+        public void Count_HasResult()
+        {
+            var context = new CollectionContext(TestSetup.Collection);
+
+            var beers = from beer in context.Query<Beer>()
+                where beer.Type == "beer"
+                select new {beer.Name};
+
+            Console.WriteLine(beers.Count());
+        }
+
+        [Test]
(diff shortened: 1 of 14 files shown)


Src/Couchbase.Linq.IntegrationTests/CountQueryTests.cs
```diff
@@ -0,0 +1,89 @@
...
+        public void Count_HasResult()
+        {
+            var context = new CollectionContext(TestSetup.Collection);
+
+            var beers = from beer in context.Query<Beer>()
+                where beer.Type == "beer"
+                select new {beer.Name};
+
+            Console.WriteLine(beers.Count());
+        }
```
jeffrymorris, line 23: Shouldn't we be asserting here (and other places)?

Src/Couchbase.Linq.IntegrationTests/CountQueryTests.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            var context = new CollectionContext(TestSetup.Collection);
+
+            var beers = from beer in context.Query<Beer>()
+                where beer.Type == "beer"
+                select new {beer.Name};
+
+            var result = await beers.CountAsync(p => p.Name == "21A IPA");
+
+            Console.WriteLine(result);
```
jeffrymorris, line 48: here

Src/Couchbase.Linq.IntegrationTests/CountQueryTests.cs
brantburnett: I'm not sure what to assert on? If the query fails, we'll get an exception. But asserting on the count seems fraught with fragility, one extra document and integration tests fail. Note: these are integration, not unit tests.

Do you have a suggestion on what assertion to add?

jeffrymorris:
Thanks Brant

PR #315, opened 2020-06-15
Add support for AnyAsync and AllAsync

brantburnett:
Motivation
----------
Support asynchronous queries which test a predicate against query
results.

Modifications
-------------
Create extension methods, expression nodes, and result operators for
AnyAsync and AllAsync.

Add tests for the async methods and reorganize tests for the sync
methods.

Results
-------
AnyAsync or AllAsync may be called on IQueryable and they function
correctly so long as the IQueryable is from a CollectionContext.

Relates to #281


jeffrymorris:
Thanks Brant

PR #316, opened 2020-06-16
Add support for SumAsync

brantburnett:
Motivation
----------
Support asynchronous queries which return sums.

Modifications
-------------
Create extension methods, an expression node, and a result operator to
handle SumAsync.

Add tests for the async methods.

Results
-------
SumAsync may be called on IQueryable and will function correctly so long
as the IQueryable is from a CollectionContext.

Relates to #281


jeffrymorris:
Thanks!

Issue #324, opened 2020-07-04
Use cluster scoping for shared services


brantburnett:
The current singleton approach for shared services like IDocumentFilterManager is not a great approach. It makes it difficult for consumers to unit test, and prevents different clusters from having different configurations within the same process.

Once (and if) http://review.couchbase.org/c/couchbase-net-client/+/131883 is merged, we can look at improving the approach using cluster scoped services.

- [X] DocumentFilterManager
- [ ] Serialization providers for enums, DateTimes, and customization

PR #329, opened 2020-07-26
Move DocumentFilterManager into cluster DI container

brantburnett:
Motivation
----------
Make the DocumentFilterManager scoped to the cluster instead of a
static singleton. This will better support multiple clusters which may
have different requirements, as well as unit testing injection.

Modifications
-------------
Rewrite DocumentFilterManager to be an instance class instead of a
singleton. Perform some cleanup as well, such as making
DocumentFilterSet truly immutable, using nullable ref types, and fixing
up the XML doc comments.

Create extensions for ClusterOptions that register Linq2Couchbase on the
cluster services and allow extensibility.

Update some of the documentation with a modern example

Results
-------
DocumentFilterManager now lives on the ICluster.ClusterServices for
each cluster, with independent configuration.

To use Linq2Couchbase, AddLinq *must* be called on ClusterOptions prior
to connecting to the cluster.

Relates to #324

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,69 +1,75 @@
-Linq2Couchbase
-==================
+# Linq2Couchbase
 
 [![Join the chat at https://gitter.im/couchbaselabs/Linq2Couchbase](https://badges.gitter.im/couchbaselabs/Linq2Couchbase.svg)](https://gitter.im/couchbaselabs/Linq2Couchbase?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-[![Build status](https://ci.appveyor.com/api/projects/status/urml45drbj7781it?svg=true)](https://ci.appveyor.com/project/Couchbase/linq2couchbase)
+![.NET Core](https://github.com/couchbaselabs/Linq2Couchbase/workflows/.NET%20Core/badge.svg)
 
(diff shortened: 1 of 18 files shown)


brantburnett:
@mgroves I'd also like your opinion on this PR, in terms of client usability. I think it's a lot more extensible and functional long-term, but it does carry the downside that AddLinq must be called on ClusterOptions when connecting to the cluster. I think it's worth it, given how it meshes with the .NET Core style, but welcome second opinions. See the updated README.md in the PR for an example.

README.md
```diff
@@ -1,69 +1,75 @@
...
+        options.Password = "password";
 
-    var context = new BucketContext(ClusterHelper.GetBucket("travel-sample"));
+        options.AddLinq();
+    });
+
+    // Wait until the cluster is bootstrapped
+    await cluster.WaitUntilReadyAsync(TimeSpan.FromSeconds(10));
+
+    var context = new BucketContext(await cluster.BucketAsync("travel-sample"));
```
jeffrymorris, line 48: Should Cluster have an extension method that creates BucketContexts? 

`cluster.GetBucketContext(string name)` or `cluster.GetQueryContext(string bucketName)`?

Open to counter-ideas here.

brantburnett: That could be interesting. I'd probably go with CreateBucketContext to make clear it's making a new one. I'd also think an extension on IBucketProvider from the DI extensions would be good, too. Feels like a separate PR, though.

jeffrymorris: @brantburnett - this ready for review?

brantburnett: I think so, up to you if you want me to add more to it now or in a separate PR.

PR #352, opened 2021-08-03
Use JSON-style string escaping

### LINKED ISSUE #348: ParsingFailureExceptions thrown when creating brewery_id index in Couchbase.Linq.IntegrationTests
While upgrading the CouchbaseNetClient dependency to 3.2.0, a number of integration tests (JoinTests_InnerJoin_IndexJoin, JoinTests_LeftJoin_IndexJoin, etc) are failing with ParsingFailureExceptions. For example the request:

```
{"statement":"CREATE INDEX brewery_id ON beer-sample(brewery_id) USING GSI WITH {\"defer_build\":False};","client_context_id":"07f299f0-0ac5-4ddd-bf2c-aac64fb3e777"}
```
Fails with the following response:

```
{
"requestID": "127a687a-bd85-4db5-8437-8a87e2f4a1d1",
"clientContextID": "07f299f0-0ac5-4ddd-bf2c-aac64fb3e777",
"errors": [{"code":3000,"msg":"syntax error - at -"}],
"status": "fatal",
"metrics": {"elapsedTime": "578.8µs","executionTime": "468.4µs","resultCount": 0,"resultSize": 0,"serviceLoad": 2,"errorCount": 1}
}
```

The issue appears to be that the bucket name is not being escaped with back ticks: "beer-sample" should be ``"`beer-sample`"``. This can be confirmed by running the query in the request in the query console in CB.

Note this happens for 3.0.7 and for 3.2.0 versions and I suspect all 3.X versions.

@brantburnett shouldn't this escaping be happening by the `N1QlHelpers.EscapeIdentifier`: https://github.com/couchbaselabs/Linq2Couchbase/blob/master/Src/Couchbase.Linq/QueryGeneration/N1QLQueryModelVisitor.cs#L137

Note that you may have to drop brewer_id index first: DROP INDEX brewery_id on `beer-sample

Jeff

brantburnett:
Motivation
----------
The current approach to string escaping causes problems if a backslash
is included, and also can't handle special characters like tab or CR.

Modifications
-------------
Serialize string and character constants to the N1QL query using a JSON
serializer.

Also change some other locations rendering strings to use double quotes
for consistency.

Results
-------
String and character constant expressions now support the full set of
characters supported by JSON.

Strings now appear on the query wrapped in double quotes instead of
single quotes.

Fixes #348

### DIFF
--- a/Src/Couchbase.Linq.UnitTests/QueryGeneration/AggregateTests.cs
+++ b/Src/Couchbase.Linq.UnitTests/QueryGeneration/AggregateTests.cs
@@ -492,7 +492,7 @@ namespace Couchbase.Linq.UnitTests.QueryGeneration
                 "FROM `default` as `Extent1` " +
                 "INNER JOIN `default` as `Extent2` ON (`Extent1`.`brewery_id` = META(`Extent2`).id) " +
                 "GROUP BY `Extent2`.`name` " +
-                "HAVING (`Extent2`.`name` >= 'N')";
+                "HAVING (`Extent2`.`name` >= \"N\")";
 
             var n1QlQuery = CreateN1QlQuery(mockBucket.Object, query.Expression);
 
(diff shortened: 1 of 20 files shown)


jeffrymorris (approved):

jeffrymorris:
Thanks Brant!

PR #360, opened 2021-11-18
Added support of DateTimeOffset in N1QlFunctions

kapral:
Adding overloads of `N1QlFunctions` which work with `DateTime` to support `DateTimeOffset` too.

I couldn't find how test data should be set up for functional tests, so assuming it's done manually.


brantburnett (approved):
Thanks for the submission, looks good!

kapral:
@brantburnett Thanks for merging this!

Any insight on when a new nuget package goes out? We need this and collections support too.

jeffrymorris:
@kapral 

Were planning on releasing a new NuGet package in the next couple of weeks, likely the 2nd week of Dec.

Jeff

PR #366, opened 2023-05-26
Fix Crash in .net 4.8

mdegroux:


brantburnett:
@mdegroux Thank you so much for the analysis and even more for giving us a PR for it. I'm not in charge of cutting releases but I'll prod in that direction.

mdegroux:
Tanks keep me informed of the build please 🙏

Issue #371, opened 2023-10-18
Issue when querying dates


mdegroux:
Hi,

We found an issue when we query dates on computer whose local is not en-us
In france, the locale is fr-fr and the decimal separator is not a . (dot) but a , (comma)

Consider the document class 
```csharp
    public class MyDocument
    {
        public DateTime SnapshotTimeUtc { get; set; }
    }
```

Consider the linq query
```csharp
IQueryable<MyDocument> query = context.Query<MyDocument>().Where(c => c.snapshotTimeUtc < DateTime.Now.AddDays(-1));
```

When we execute it, it throws the exception

> Couchbase.Core.Exceptions.ParsingFailureException
>   HResult=0x80131500
>   Message=syntax error - line 1, column 138, near 'tc`) < 1697533605453', at: , [3000]
>   Source=Couchbase.NetClient
>   StackTrace:
>    at Couchbase.Query.QueryResultExtensions.ThrowExceptionOnError[T](IQueryResult`1 result, QueryErrorContext context) in C:\temp\couchbase-net-client-master\src\Couchbase\Query\QueryResultExtensions.cs:line 137
>    at Couchbase.Query.QueryClient.<ExecuteQuery>d__14`1.MoveNext() in C:\temp\couchbase-net-client-master\src\Couchbase\Query\QueryClient.cs:line 258
>    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
>    at Couchbase.Query.QueryClient.<QueryAsync>d__13`1.MoveNext() in C:\temp\couchbase-net-client-master\src\Couchbase\Query\QueryClient.cs:line 90
>    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
[...]

PR #372, opened 2023-10-18
Fix datetime queries with non us locales

mdegroux:
description in issue #371


brantburnett:
@mdegroux Thanks for the find and the fix. I did one tweak to the PR to avoid allocating an intermediate string on the heap in .NET 6. Check it out and make sure you think it's still okay and we'll get it merged.

Thanks again.

brantburnett (approved):

mdegroux:
Didn't knew this trick
If your are confident with, it's ok

PR #374, opened 2024-08-07
Added DateMethodCallTranslator to better support Date truncation

gritse:
We can now use the `DateTime.Date` and `DateTimeOffset.Date` properties in LINQ without needing the `N1QlFunctions.DateTrunc` method. This enhancement makes LINQ expressions cleaner and more consistent with other LINQ providers that already support this behavior.

### DIFF
--- a/Src/Couchbase.Linq/QueryGeneration/MethodCallTranslators/DateMethodCallTranslator.cs
+++ b/Src/Couchbase.Linq/QueryGeneration/MethodCallTranslators/DateMethodCallTranslator.cs
@@ -1,0 +1,41 @@
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using System.Reflection;
+
+namespace Couchbase.Linq.QueryGeneration.MethodCallTranslators;
+
+internal class DateMethodCallTranslator : IMethodCallTranslator
+{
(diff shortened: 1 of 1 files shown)


Src/Couchbase.Linq/QueryGeneration/MethodCallTranslators/DateMethodCallTranslator.cs
```diff
@@ -0,0 +1,41 @@
...
+            throw new ArgumentNullException("methodCallExpression");
+        }
+
+        var expression = expressionTreeVisitor.Expression;
+
+        expression.Append("DATE_TRUNC_STR(");
+        expressionTreeVisitor.Visit(methodCallExpression.Object);
+        expression.Append(",");
+        expression.Append(@"""day""");
+        expression.Append(")");
```
brantburnett, line 37: It would be slightly more performant to make this a single append operation.

```suggestion
        expression.Append(@",""day"")");
```

brantburnett (changes requested):
Thanks for the submission, it looks good! Just one minor point.

brantburnett:
Thanks again for the submission, I merged it and added some additional tests around it.

Is this something that's blocking a project you're working on? @jeffrymorris manages the releases for this repo, not me.

PR #378, opened 2024-11-09
Apply document filters to DocumentSet<T> at query time

### LINKED ISSUE #376: Support for `DocumentFilter` with Custom Bucket Context `IDocumentSet`
Description
I've noticed that the DocumentFilter functionality currently only works with the BucketContext.Query method and does not seem to apply when using an IDocumentSet within a custom bucket context.

Use Case
In my application, I am working with a custom bucket context where I use IDocumentSet. It would be very helpful to be able to apply DocumentFilter directly on this IDocumentSet to take advantage of consistent filtering across different bucket contexts.

Expected Behavior
I expected DocumentFilter to be compatible with both BucketContext.Query and IDocumentSet, allowing seamless filtering regardless of the context.

Current Behavior
Currently, DocumentFilter only works with BucketContext.Query, making it challenging to use in custom bucket contexts that rely on IDocumentSet.

brantburnett:
Motivation
----------
Some forms of querying using `DocumentSet<T>` properties on an inherited `BucketContext` don't apply document filters such as the `DocumentType` attribute.

Modifications
-------------
- Split `CollectionQueryable<T>` into two types `CollectionQueryable<T>` and a base `CouchbaseQueryable<T>`. The former is used to represent the original extent of a query against a collection. The latter is used to construct new `IQueryable<T>` instances as the query is extended with predicates, etc. This removes unnecessary fields and logic from the simpler case that is repeated for every new LINQ method applied to the query.
- Add an additional non-generic `IDocumentSet` interface which is inherited by `IDocumentSet<T>`. This is a SemVer safe change because it only adds `IQueryable` which was already included on `IDocumentSet<T>` via `IQueryable<T>`.
- Create DelayedFilterQueryProvider as a wrapper for ClusterQueryProvider that applies filters to `IDocumentSet` at query execution time.
- Refactor query timeouts to always be based on a callback to the `BucketContext` rather than a property, which allows a singleton `ClusterQueryExecutor` per `BucketContext` rather than per query.
- Refactor `BucketContext` to build a singleton query provider, query parser, and query executor rather than recreating for every query.
- Refactor `DocumentSet<T>` to get the query provider from the `BucketContext`.
- Reduce interface invocations by caching the bucket, scope, and collection names once when constructing a `DocumentSet<T>` or `CollectionQueryable<T>` instead of for each property read.
- Minor perf and nullable ref type improvements to `DocumentFilterSet`.
- Fix some integration tests that were refering to data no longer found in the default `beer-sample` bucket.

Results
-------
Filters are applied consistently to `DocumentSet<T>` based on the filter configuration at the time the query is run. Queries in all cases will have fewer heap allocations and other CPU performance benefits.

Resolves #376

### DIFF
--- a/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
+++ b/Src/Couchbase.Linq.IntegrationTests/QueryTests.cs
@@ -174,7 +174,7 @@ namespace Couchbase.Linq.IntegrationTests
             var context = new BucketContext(TestSetup.Bucket);
 
             var beers = from b in context.Query<Beer>()
-                        where b.Type == "beer" && b.Name.StartsWith("563")
+                        where b.Type == "beer" && b.Name.StartsWith("Amendment")
                         select new { name = b.Name, abv = b.Abv };
 
             var results = beers.Take(1).ToList();
@@ -1000,7 +1000,7 @@ namespace Couchbase.Linq.IntegrationTests
             var context = new BucketContext(TestSetup.Bucket);
 
             var breweries = from brewery in context.Query<Brewery>()
-                where brewery.Type == "brewery" && brewery.Address.Contains("563 Second Street")
+                where brewery.Type == "brewery" && brewery.Address.Contains("210 Aberdeen Dr.")
                 orderby brewery.Name
                 select new {name = brewery.Name, addresses = brewery.Address};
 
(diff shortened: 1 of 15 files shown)


jeffrymorris (approved):
Looks good, thanks Brant!

PR #380, opened 2024-11-11
Target .NET 8

brantburnett:
Motivation
----------
General modernization.

Modifications
-------------
- Add global.json to require the .NET 8 SDK
- Use LangVersion 12 in Directory.Build.props
- Drop the deprecated .NET Standard 2.1 build target and add .NET 8
- Update to Couchbase SDK 3.5.5
- Upgrade various NuGet dependencies, especially for unit tests, and drop some unnecessary dependencies
- Use the in-box SourceLink in .NET 8 SDK rather than the OOB package
- Target net8.0 for unit tests
- Fix a LOT of nullable reference type warnings due to improved annotations in .NET 8


jeffrymorris (approved):
Looks good, thanks!

PR #385, opened 2025-06-26
Make Query<T>(BucketQueryOptions options) virtual to allow customizat…

d5dl2:
Added the virtual keyword to the Query<T>(BucketQueryOptions options) method in BucketContext, enabling developers to override it for custom scope and collection name handling.


brantburnett (changes requested):
Thanks for the submission, but I'm not really a fan of making this virtual. While it does meet the need, it raise other long term maintenance risks. As soon as you start making methods virtual it starts adding a lot of complexity in the future when you want to make other design changes. The .NET team ran into a lot of similar issues with all the virtual members on `System.IO.Stream`, for example.

I think we should consider other methods for extensibility, rather than simple virtual methods. A common pattern used by the .NET team these days is to make a protected virtual `XXXCore` method that takes a `class` or `struct` as the parameter. Then make the non-virtual public methods build the parameter object and pass it to the core method. Since the parameter is an object, it can safely receive more parameters in the future without it being a binary breaking change.

For example, something like this:

```c#
protected struct QueryRequest
{
    public string Scope { get; set; }
    public string Collection { get; set; }
    public BucketQueryOptions Options { get; set; }
}

public IQueryable<T> Query<T>(BucketQueryOptions options) =>
    QueryCore<T>(new QueryRequest { /* ... */);

internal IQueryable<T> Query<T>(string scope, string collection, BucketQueryOptions options) =>
    QueryCore<T>(new QueryRequest { /* ... */);

protected virtual IQueryable<T> QueryCore<T>(QueryRequest request)
{
    // Move logic from the second overload of Query here
}
```

jeffrymorris:
Closing this as no more work has been done on it since it was reviewed.

